PR #4, opened 2016-03-19
Removes an article from Instant Articles library

scottrigby:
Hey guys, would this be a valuable method for `Facebook\InstantArticles\Client`? I know it's basically a wrapper for the graph API, but this lets us easily make use of the `getArticleIDFromCanonicalURL()` method, and may be nice to have an easy wrapper.

Anyway, if you think it makes sense, should we throw an exception if the article URL doesn't exist?

### DIFF
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -102,6 +102,10 @@ class Client
      *
      * @param $canonicalURL
      * @return \Facebook\InstantArticles\Client\InstantArticleStatus
+     *
+     * @todo Consider returning the \Facebook\FacebookResponse object sent by
+     *   \Facebook\Facebook::delete(). For now we trust that if an Instant
+     *   Article ID exists for the Canonical URL the delete operation will work.
      */
     public function removeArticle($canonicalURL)
     {
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -117,7 +117,7 @@ class Client
 
         if ($articleID = $this->getArticleIDFromCanonicalURL($canonicalURL)) {
             $this->facebook->delete($articleID);
-            return InstantArticleStatus::success(array('An Instant Article ID ' . $articleID . ' was found for ' . $canonicalURL . ' in ' . __FUNCTION__ . '.'));
+            return InstantArticleStatus::success();
         }
         return InstantArticleStatus::notFound(array('An Instant Article ID ' . $articleID . ' was not found for ' . $canonicalURL . ' in ' . __FUNCTION__ . '.'));
     }
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -100,7 +100,7 @@ class Client
     /**
      * Removes an article from your Instant Articles library.
      *
-     * @param $canonicalURL
+     * @param string $canonicalURL The canonical URL of the article to delete.
      * @return \Facebook\InstantArticles\Client\InstantArticleStatus
      *
      * @todo Consider returning the \Facebook\FacebookResponse object sent by
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -98,6 +98,17 @@ class Client
     }
 
     /**
+     * Removes an article from your Instant Articles library.
+     *
+     * @param string $canonicalURL The canonical URL of the article to get the status for.
+     */
+    public function removeArticle($canonicalURL) {
+        if ($articleID = $this->getArticleIDFromCanonicalURL($canonicalURL)) {
+            $this->facebook->delete($articleID);
+        }
+    }
+
+    /**
      * Get an Instant Article ID on its canonical URL.
      *
      * @param string $canonicalURL The canonical URL of the article to get the status for.
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -100,7 +100,7 @@ class Client
     /**
      * Removes an article from your Instant Articles library.
      *
-     * @param string $canonicalURL The canonical URL of the article to get the status for.
+     * @param string $canonicalURL The canonical URL of the article to delete.
      */
     public function removeArticle($canonicalURL) {
         if ($articleID = $this->getArticleIDFromCanonicalURL($canonicalURL)) {
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -100,12 +100,22 @@ class Client
     /**
      * Removes an article from your Instant Articles library.
      *
-     * @param string $canonicalURL The canonical URL of the article to delete.
+     * @param $canonicalURL
+     * @return \Facebook\InstantArticles\Client\InstantArticleStatus
      */
-    public function removeArticle($canonicalURL) {
+    public function removeArticle($canonicalURL)
+    {
+        if (!$canonicalURL) {
+            return InstantArticleStatus::notFound(array('$canonicalURL param not passed to ' . __FUNCTION__ . '.'));
+        }
+
+        Type::enforce($canonicalURL, Type::STRING);
+
         if ($articleID = $this->getArticleIDFromCanonicalURL($canonicalURL)) {
             $this->facebook->delete($articleID);
+            return InstantArticleStatus::success(array('An Instant Article ID ' . $articleID . ' was found for ' . $canonicalURL . ' in ' . __FUNCTION__ . '.'));
         }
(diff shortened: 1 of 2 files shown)


scottrigby:
OK I signed the Contributor License Agreement. Should be all good now.

scottrigby:
Regarding exceptions, it looks like `\Facebook\FacebookClient::sendRequest()` should already throw errors if needed.

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -98,6 +98,17 @@ public function importArticle($article, $takeLive = false)
     }
 
     /**
+     * Removes an article from your Instant Articles library.
+     *
+     * @param string $canonicalURL The canonical URL of the article to get the status for.
```
simonengelhardt: *delete

simonengelhardt: We should add a `@return InstantArticleStatus` (see comment below)

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -98,6 +98,17 @@ public function importArticle($article, $takeLive = false)
...
+     */
+    public function removeArticle($canonicalURL) {
+        if ($articleID = $this->getArticleIDFromCanonicalURL($canonicalURL)) {
```
simonengelhardt: I think it would be a good idea to add a check identical to [this one](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Client/Client.php#L131) and return an `InstantArticleStatus` instance – both for "not found" and "success"

We should also add at type check like [this one](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Client/Client.php#L135).

simonengelhardt:
Thanks for putting this together, @scottrigby. I've added a few inline comments.

src/Facebook/InstantArticles/Client/Client.php
scottrigby: Regarding `InstantArticleStatus`, I was wondering why you do a custom not found status, since PHP will already issue an error if you don't pass a param (if we had to do this for every param without defaults for every function in PHP it would get pretty cumbersome). But if that's a standard you use for some other purpose it's your SDK so I will add it. Just curious why this is done.

scottrigby:
Regarding `InstantArticleStatus`, I was wondering why you do a custom not found status, since PHP will already issue an error if you don't pass a param (if we had to do this for every param without defaults for every function in PHP it would get pretty cumbersome). But if that's a standard you use for some other purpose it's your SDK so I will add it. Just curious why this is done.

scottrigby:
@simonengelhardt  Regarding `Type::enforce` I'm happy to add a custom type check, but I didn't because `getArticleIDFromCanonicalURL()` which this method calls already does `Type::enforce($canonicalURL, Type::STRING);`. Do we really need to do it again? Either way let me know and i will follow your SDK standards.

scottrigby:
@simonengelhardt One other note: I noticed `\Facebook\Facebook::delete()` ultimately returns a `\Facebook\FacebookResponse` object (via `\Facebook\Facebook::sendRequest()`). I didn't do anything with that response in this new method because I was following the lead from `\Facebook\InstantArticles\Client\Client::importArticle()` (which just calls `post()` and doesn't check or return any status from that). I wonder if we should do any of that with both of these methods? It could also be a follow-up :)

simonengelhardt:
Thanks for the feedback, @scottrigby.

`InstantArticleStatus` wraps the object returned from a status call to the API, which includes errors and warnings from the ingestion. We extended that with a "not found" status, so that it would be easy for the developer using `Client` to create UI for showing status in a CMS. I think it makes sense to reuse that status type for this method, now that we have it.

I think it would be best to add a `Type::enforce` call directly in this method. This is essentially in lieu of type declarations in the method signature, which we can't really use in PHP >=5.4

About the return, maybe it could be useful to return the `Facebook\FacebookResponse` object at some point, but we can add that separately. The thing is that both import and delete are async on the server, so the returned response can't say anything about the outcome of the operation itself. You will need to query `getLastSubmissionStatus` for that.

scottrigby:
@simonengelhardt OK cool. thanks for the larger picture. I'll update this on my next free moment :smile:

scottrigby:
@simonengelhardt I drew this from the examples you pointed to, plus what you asked for in the now hidden comments above (I also clarified what `notFound` and `success` means in the optional `$messages` param. Additionally I added a Doxygen `@todo` for the `FacebookResponse` since that's outside the scope of this PR (and applies to all the methods if we implement that here). This all may be a little verbose, so let me know if that's not in line with how you want the SDK to be written, and if that's the case I can take that down a notch :smile:.

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -98,6 +98,31 @@ public function importArticle($article, $takeLive = false)
...
+    {
+        if (!$canonicalURL) {
+            return InstantArticleStatus::notFound(array('$canonicalURL param not passed to ' . __FUNCTION__ . '.'));
+        }
+
+        Type::enforce($canonicalURL, Type::STRING);
+
+        if ($articleID = $this->getArticleIDFromCanonicalURL($canonicalURL)) {
+            $this->facebook->delete($articleID);
+            return InstantArticleStatus::success(array('An Instant Article ID ' . $articleID . ' was found for ' . $canonicalURL . ' in ' . __FUNCTION__ . '.'));
```
diegoquinteiro: We don't need a message for the success case as it is auto-explanatory, you can simply use `InstantArticleStatus::success()` with no args.

diegoquinteiro:
Thanks for that @scottrigby, the code looks great to me.

Can you add a unit test to it before merging? There are tests for the other methods, you can just use them as an example.

Take a look at my inline comment too.

scottrigby:
Hey @diegoquinteiro it was great to meet up at the camp a while back :smile: 

OK so what is your recommended way to get the canonical URL in methods on `\Facebook\InstantArticles\Client\ClientTest` so that I can test removing the article? Should I repeat the guts of other test functions, or can I reuse them… something like this?:
```diff
diff --git a/tests/Facebook/InstantArticles/Client/ClientTest.php b/tests/Facebook/InstantArticles/Client/ClientTest.php
index dde4e8c..c53fc7f 100644
--- a/tests/Facebook/InstantArticles/Client/ClientTest.php
+++ b/tests/Facebook/InstantArticles/Client/ClientTest.php
@@ -64,6 +64,20 @@ class ClientTest extends \PHPUnit_Framework_TestCase
         $this->client->importArticle($this->article, true);
     }
 
+    public function testRemoveArticle()
+    {
+        $canonicalURL = "http://facebook.com";
+        $this->article->withCanonicalURL($canonicalURL);
+        $this->testImportArticle();
+
+        $this->facebook
+            ->expects($this->once())
+            ->method('delete')
+            ->with($canonicalURL);
+
+        $this->client->removeArticle($canonicalURL);
+    }
+
     public function testImportArticleDevelopmentMode()
     {
         $this->client = new Client(
```

scottrigby:
@diegoquinteiro Also, do you ideally want to add [PHPUnit annotations](https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.coversDefaultClass) like `@beforeClass` and `@covers`?

diegoquinteiro:
Hi @scottrigby! Yes, it was nice to meet you there, and I'm excited you are helping us with the SDK =)

I don't have a strong opinion on how to create the test, but I usually do not reuse other test methods to avoid coupling. My only concern is that, doing like you said, the assertions on 'testImportArticle' would run on the 'testDelete'.

If you wanna reuse some code, you can extract the part you wanna share to another method, what do you think?

About the annotations, we didn't provide them for other tests but if you wanna include on yours it'll be nice =)

diegoquinteiro:
What about?

```php
    public function testRemoveArticle()
    {
        $canonicalURL = 'http://facebook.com';
        $articleID = '1';

        // Use a mocked client with stubbed getArticleIDFromCanonicalURL()
        $this->client = $this->getMockBuilder('Facebook\InstantArticles\Client\Client')
            ->setMethods(array('getArticleIDFromCanonicalURL'))
            ->setConstructorArgs(array(
                $this->facebook,
                "PAGE_ID",
                true // developmentMode
            ))->getMock();

        $this->client
            ->expects($this->once())
            ->method('getArticleIDFromCanonicalURL')
            ->with($canonicalURL)
            ->will($this->returnValue($articleID));;

        $this->facebook
            ->expects($this->once())
            ->method('delete')
            ->with($articleID);

        $this->client->removeArticle($canonicalURL);
    }
```

scottrigby:
@diegoquinteiro this looks perfect, and thanks for taking the time to look at this with me offline. I updated the PR.

For anyone else (who like me are not fully familiar with the testing framework yet, what this does is forward all methods to the original class, except for the one specified. This method is replaced with one that:

1. Returns the value "1" when called with the specified URL.
2. Expects to be called once with this URL.
The reason for doing this is so we can test a single method without needing to rely on it's dependencies (getArticleIDFromCanonicalURL).

diegoquinteiro:
:shipit:

PR #13, opened 2016-04-03
Fix type hints in the doc block for Elements.

m4olivei:
While working with the API in PHPStorm, I noticed that the doc blocks for some Element::create() methods were missing in such a way that auto-complete wouldn't work with those Element objects.


simonengelhardt:
Thanks! 👍

PR #24, opened 2016-04-11
fix AnimatedGif class references

mfelser:
correct capitalization fixing class not found errors


diegoquinteiro:
Hi @mfelser, thank you so much for putting this together! 

I actually went the other way around and made 'AnimatedGIF' the standard. 

That's already on master and I've also released v1.0.1 including the fix. 

Can you check if that fixes this issue?

diegoquinteiro:
If it does not, please open a issue.

Thanks!

PR #37, opened 2016-04-15
Fix: No need to specify an array of paths when there's one only

localheinz:
This PR

* [x] specifies a single path for PSR-4 autoloading instead of an array containing that path

Somewhat related to #36.


simonengelhardt:
Thanks 👍

PR #38, opened 2016-04-15
Fix: Limit range of accepted PHP versions

localheinz:
This PR

* [x] limits the range of acceptable PHP versions in `composer.json`

🙎 While PHP8 isn't out yet, it's not clear whether this package will be compatible with PHP8 or even PHP9000, right?


simonengelhardt:
That is completely right 😄 

Merging...

localheinz:
@simonengelhardt 

😃

PR #39, opened 2016-04-15
Fix: Remove unused imports

localheinz:
This PR

* [x] removes unused imports

Follows #17.

🙎 Of course, like one does, using [`fabpot/php-cs-fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and running

```
$ php-cs-fixer fix --rules=no_unused_imports src
$ php-cs-fixer fix --rules=no_unused_imports tests
```


everton-rosario:
Thanks for the PR!
@localheinz you are pushing so much good stuff, improving and making this SDK even better!

localheinz:
@everton-rosario 

We've built our own implementation at @refinery29, looking into using this, so happy to contribute!

PR #43, opened 2016-04-15
Fix: Postion vs Position

localheinz:
This PR

* [x] deprecates misspelled methods `withPostion()` and delegates to one that is correctly spelled instead

### DIFF
--- a/src/Facebook/InstantArticles/Elements/H2.php
+++ b/src/Facebook/InstantArticles/Elements/H2.php
@@ -65,15 +65,27 @@ class H2 extends TextContainer
     }
 
     /**
-    * The Text position that will be used.
-    *
-    * @see Caption::POSITION_ABOVE
-    * @see Caption::POSITION_BELOW
-    * @see Caption::POSITION_CENTER
-    *
-    * @param string position that will be used.
-    */
+     * @deprecated
+     *
+     * @param string $position
+     * @return $this
+     */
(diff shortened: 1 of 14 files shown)


src/Facebook/InstantArticles/Elements/H2.php
```diff
@@ -65,16 +65,28 @@ public function withTextAlignment($text_alignment)
     }
 
     /**
-    * The Text position that will be used.
```
localheinz: @simonengelhardt 

Unrelated to this PR, but there are a lot of docblocks which aren't indented correctly.

everton-rosario:
@localheinz, this was a nice catch.
Unfortunately it happens in much more files that had this misspelling copy/pasted.

Can you check those other files as well?
Elements with misspelled method "withPostion"
- Caption.php
- Cite.php
- H1.php
- H2.php

Usages of elements
- CaptionCreditRule.php
- CaptionRule.php
- H1Rule.php
- H2Rule.php
- CaptionTest.php
- CiteTest.php
- H1Test.php
- H2Test.php
- ImageTest.php
- MapTest.php

localheinz:
@everton-rosario 

Give me a moment!

localheinz:
@everton-rosario 

👍

everton-rosario:
Thanks for this @localheinz

PR #44, opened 2016-04-15
Fix: Use appropriate annotation for return types

localheinz:
This PR

* [x] uses `@return` instead of `@param` for annotating return types in accessors


everton-rosario:
Thanks for this fix @localheinz

PR #46, opened 2016-04-16
Refactor and optimize transformer up to 10x faster

diegoquinteiro:
I've changed the internals of the transformer so it performs over 10x faster on the complete example. Changes were commented at the code.

Also removed some arbitrary logging that are leftover from our internal testing.

The outside interface was kept compatible, but I've fixed `getRules()` to return the rules in the order they were inserted, making more sense as now `getRules()` will return exactly the value set using `setRules()`.

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Rules/Rule.php
+++ b/src/Facebook/InstantArticles/Transformer/Rules/Rule.php
@@ -16,6 +16,24 @@ abstract class Rule
 
         $matches_context = $this->matchesContext($context);
         $matches_node = $this->matchesNode($node);
+        if ($matches_context && $matches_node) {
+            $log->debug('context class: '.get_class($context));
+            $log->debug('context matches: '.($matches_context ? 'MATCHES' : 'no match'));
+            $log->debug('node name: <'.$node->nodeName.' />');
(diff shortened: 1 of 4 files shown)


src/Facebook/InstantArticles/Transformer/Rules/Rule.php
```diff
@@ -16,24 +16,6 @@ public function matches($context, $node)
...
-        }
-        if ($node->nodeName === 'img') {
-            $log->debug('context class: '.get_class($context));
-            $log->debug('context matches: '.($matches_context ? 'MATCHES' : 'no match'));
-            $log->debug('node name: <'.$node->nodeName.' />');
-            $log->debug('node: '.$node->ownerDocument->saveXML($node).' />');
-            $log->debug('node matches: '.($matches_node ? 'MATCHES' : 'no match'));
-            $log->debug('rule: '.get_class($this));
-            $log->debug('-------');
-        }
```
everton-rosario: Hey @diegoquinteiro, please, do not merge 2 things on same pull request.
This should not be lost unless it is addressed in another way.
This logging information is very helpful when trying to debug the transformation of a document.

PR #47, opened 2016-04-16
Fix: Withing should probably be Within

localheinz:
This PR

* [x] fixes a typo within a method name

:trollface:


everton-rosario:
Yes, totally makes sense.

Thanks for the PR!

PR #48, opened 2016-04-16
Fix: Update symfony/css-selector

### LINKED ISSUE #45: Installation fails, requires older version of syfony/css-selector
I have installed symfony/css-selector v 3.0.4 and when I try to install this package it says that it requires  symfony/css-selector 2.8.* and thus it won't install. I can't downgrade symfony/css-selector because it's required for another package. Is this something that can be fixed?

localheinz:
This PR

* [x] updates `symfony/css-selector`

Fixes #45.

💁 For reference, see https://github.com/symfony/css-selector/compare/v2.8.4...v3.0.4.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,9 @@
     }],
     "require": {
         "php": ">=5.4",
-        "symfony/css-selector": "2.8.*",
+        "apache/log4php": "2.3.0",
         "facebook/php-sdk-v4": "~5.0",
-        "apache/log4php": "2.3.0"
+        "symfony/css-selector": "^3.0.4"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8"
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -11,9 +11,9 @@
     }],
     "require": {
         "php": ">=5.4",
-        "symfony/css-selector": "2.8.*",
+        "apache/log4php": "2.3.0",
         "facebook/php-sdk-v4": "~5.0",
-        "apache/log4php": "2.3.0"
+        "symfony/css-selector": "^3.0.4"
```
localheinz: Just a heads-up, always using the `--sort-packages` option, since it is a lot easier to spot items in a list when that list is sorted.

```
$ composer require --sort-packages symfony/css-selector:^3.0.4
```

For reference, see https://github.com/composer/composer/pull/3549.

everton-rosario: Hello @localheinz, thanks for your PR!

Why have you moved symfony/css-selector from 2.8.* to 3.0.4?
`"symfony/css-selector": "^3.0.4`
We downgraded it just because 2.8.* is compatible with PHP 5.4

The order of packages makes sense, but we are not intended in not being compatible with PHP 5.4 because of its usage nowadays.

Can you change the version back to 2.8.*?

localheinz: @everton-rosario 

Ha, wasn't aware of it, closing, then!

PR #51, opened 2016-04-18
Localheinz fix/bootstrapping

everton-rosario:
Solved the conflict and merged PR of @localheinz

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "homepage": "https://github.com/facebook/facebook-instant-articles-sdk-php/contributors"
     }],
     "require": {
-        "php": ">=5.4",
+        "php": "^5.4 || ^7.0",
         "symfony/css-selector": "2.8.*",
         "facebook/php-sdk-v4": "~5.0",
         "apache/log4php": "2.3.0"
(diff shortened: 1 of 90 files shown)


localheinz:
@everton-rosario 

I had resolved the conflict on https://github.com/facebook/facebook-instant-articles-sdk-php/pull/36 already, though. 😄

everton-rosario:
We need to add composer.lock to .gitignore
Can you send a PR about it, @localheinz ?

simonengelhardt:
I don't think we should gitignore composer.lock since that goes against the [recommendation from Composer](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file) and could potentially break our upcoming CI integration.

localheinz:
@everton-rosario @simonengelhardt 

Personally, I would recommend to always check in `composer.lock` as it is a lot easier to reproduce issues, which otherwise could be caused by using different versions of dependencies.

However, major frameworks often don't check it in, for example

* https://github.com/symfony/symfony/blob/master/.gitignore#L2
* https://github.com/zendframework/zf2/blob/master/.gitignore#L12

How about checking in `composer.lock`, though, as it speeds up development, and then potentially running at least those builds:

* one that uses the lowest dependencies (`composer update --prefer-lowest`)
* one that uses locked dependencies (`composer install`)
* one that uses the highest dependencies (`composer update`)

For reference, see

* https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file
* https://getcomposer.org/doc/03-cli.md#update

PR #53, opened 2016-04-18
Fix: Run phpcs on Travis

localheinz:
This PR

* [x] requires [`squizlabs/php_codesniffer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and updates `CONTRIBUTING.md` accordingly
* [x] provides a configuration for `phpcs` and `phpcbf`
* [x] runs `phpcs` on Travis (only on one build, though)
* [x] runs `phpcbf` to fix coding style issues

PR #54, opened 2016-04-18
Fix: Simplify asserting that InvalidArgumentException is thrown

localheinz:
This PR

* [x] simplifies tests asserting that `InvalidArgumentException`s are thrown

💁 Spotted in #53.


tests/Facebook/InstantArticles/Elements/Validators/TypeTest.php
```diff
@@ -360,11 +329,8 @@ public function testEnforceWithinTrueString()
 
     public function testEnforceWithinExceptionString()
     {
-        try {
-            $result = Type::enforceWithin('a', array('x', 'y', 'z'));
-            $this->fail('Should trhow exception');
```
localheinz: Here's a typo, by the way.

PR #56, opened 2016-04-18
Fix: Use willReturn()

localheinz:
This PR

* [x] uses `willReturn($value)` instead of `will($this->returnValue($value))` when setting up test doubles


tests/Facebook/InstantArticles/Client/ClientTest.php
```diff
@@ -87,7 +87,7 @@ public function testRemoveArticle()
           ->expects($this->once())
           ->method('getArticleIDFromCanonicalURL')
           ->with($canonicalURL)
-          ->will($this->returnValue($articleID));;
```
localheinz: I've removed an extra `;` here, this could be done automatically using #53 and the `duplicate_semicolon` fixer, see https://github.com/FriendsOfPHP/PHP-CS-Fixer#usage:

> **duplicate_semicolon [symfony]**
> Remove duplicated semicolons.

PR #58, opened 2016-04-18
Fix: Remove unused import referencing undefined class

localheinz:
This PR

* [x] removes an unused import referencing an undefined class

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Getters/GetterFactory.php
+++ b/src/Facebook/InstantArticles/Transformer/Getters/GetterFactory.php
@@ -8,8 +8,6 @@
  */
 namespace Facebook\InstantArticles\Transformer\Getters;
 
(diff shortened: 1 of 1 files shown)


src/Facebook/InstantArticles/Transformer/Getters/GetterFactory.php
```diff
@@ -8,8 +8,6 @@
  */
 namespace Facebook\InstantArticles\Transformer\Getters;
 
-use Facebook\InstantArticles\Transformer\Validators\Type;
```
localheinz: The class [`Facebook\InstantArticles\Validators\Type`](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Validators/Type.php) exists, but this import is never used anyway.

PR #60, opened 2016-04-19
Enhancement: Cache dependencies installed with Composer between builds

localheinz:
This PR

* [x] caches dependencies installed with Composer between Travis builds

💁 For reference, see https://docs.travis-ci.com/user/caching/#Arbitrary-directories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,15 @@ php:
   - 7.0
   - hhvm
 
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 before_install:
   - travis_retry composer self-update
 
 install:
-  - travis_retry composer install --no-interaction
+  - travis_retry composer install --no-interaction --prefer-dist
 
 script:
   - vendor/bin/phpunit


.travis.yml
```diff
@@ -0,0 +1,35 @@
+language: php
+
+sudo: false
+
+env:
+  global:
+    - GITHUB_TOKEN=foo
```
localheinz: An encrypted, actual GitHub token should be added here

Run

```
$ travis encrypt GITHUB_TOKEN=...
```

and then replace this

```diff
-    - GITHUB_TOKEN=foo
+    - secure: ...
```

💁 This helps with preventing issues regarding GitHub API rate limits, when installing dependencies with `composer`.

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+matrix:
+  fast_finish: true
+  include:
+    - php: 5.4
+    - php: 5.5
+    - php: 5.6
+    - php: 7
+    - php: hhvm
+ allow_failures:
+    - php: hhvm
```
localheinz: The build for `hhvm` failed, so I added it here.

See https://travis-ci.org/facebook/facebook-instant-articles-sdk-php/builds/124255378.

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+
+before_install:
+  - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi
```
localheinz: This helps speeding up the build and should conditionally be removed when interested in code coverage, of course.

.travis.yml
simonengelhardt: I actually went ahead and fixed the tests that were breaking `hhvm` over in bf226b4062c0fb5b38fedaa7f2ae8b8d5dbabd43

simonengelhardt:
@localheinz Thank you so much for this 👍  Meanwhile, I've set up a bare-bones `.travis.yml` in #61, but will be sure to take a look at your enhancements here.

localheinz:
@simonengelhardt 

I've repurposed this PR to speed up the build by caching dependencies between builds on Travis.

PR #70, opened 2016-04-20
Fix Composer license identifier

simonengelhardt:
"Facebook Platform" is not a proper [SPDX License Identifier](https://spdx.org/licenses/). Let's follow the [Composer schema](https://getcomposer.org/doc/04-schema.md#license) and set to "proprietary". This will remove the warning when running `composer validate` (as mentioned in #67).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "keywords": ["facebook", "sdk", "instant", "articles"],
     "type": "library",
     "homepage": "https://github.com/facebook/facebook-instant-articles-sdk-php",
-    "license": "Facebook Platform",
+    "license": "proprietary",
     "authors": [{
         "name": "Facebook",
         "homepage": "https://github.com/facebook/facebook-instant-articles-sdk-php/contributors"


localheinz:
@simonengelhardt 

At https://getcomposer.org/doc/04-schema.md#license: 

> For closed-source software, you may use "proprietary" as the license identifier.

However, the content of [`LICENSE`](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/LICENSE) suggests that it should be MIT, what do you think?

For reference, see https://en.wikipedia.org/wiki/MIT_License.

simonengelhardt:
@localheinz It's definitely derived from MIT, but not exactly MIT since there is a clause in there that the software is "... for use in connection with the web services and APIs provided by
Facebook"

Since this license, that we've called "Facebook Platform", does not have an SPDX identifier, I thought "proprietary" was the only accurate choice even though it is very much open source. Do you have any better suggestion? Thanks!

localheinz:
@simonengelhardt 

Should probably work, then! 👍

PR #74, opened 2016-04-21
Fix: Use short array syntax

localheinz:
This PR

* [x] uses short array syntax

Follows #39.

🙎 Of course, like one does, using [`fabpot/php-cs-fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and running

```
$ php-cs-fixer fix --rules=short_array_syntax src
$ php-cs-fixer fix --rules=short_array_syntax tests
```


simonengelhardt:
Thanks for the housekeeping. LGTM 👍

PR #83, opened 2016-05-01
Fix: Remove unused private fields

localheinz:
This PR

* [x] removes unused `private` fields

💁 They are never assigned a value or read from.


everton-rosario:
Can you double check what is in conflict once other PR's were merged?

Sorry for the delay, I was out last week.

localheinz:
@everton-rosario 

Resolved!

PR #94, opened 2016-05-06
Enhancement: Add composer scripts to ease development

localheinz:
This PR

* [x] adds composer scripts to ease development

Run

```
$ composer cs
```

to fix and check coding style issues.

Run

```
$ composer test
```

to run the tests.

Run

```
$ composer all
```

if you are extra lazy to run both.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,5 +28,19 @@
         "psr-4": {
             "Facebook\\InstantArticles\\": "tests/Facebook/InstantArticles/"
         }
+    },
+    "scripts": {
+        "all": [
+            "@cs",
+            "@test"
+        ],
+        "cs": [
+            "composer install",
+            "phpcbf --standard=phpcs.xml -p || phpcs --standard=phpcs.xml -p"
+        ],
+        "test": [
+            "composer install",
+            "phpunit"
+        ]
     }
 }
--- a/README.md
+++ b/README.md
@@ -25,17 +25,32 @@ composer require facebook/facebook-instant-articles-sdk-php
 
 [Composer](https://getcomposer.org/) is a prerequisite for testing and developing. [Install composer globally](https://getcomposer.org/doc/00-intro.md#globally), then install project dependencies by running this command in the project root directory:
 ```sh
-composer install
+$ composer install
 ```
 
 To run the tests:
+
 ```sh
-./vendor/bin/phpunit
+$ composer test
 ```
+
+To fix and check for coding style issues:
+
+```sh
+$ composer cs
+```
+
+Extra lazy? Run
+
+```sh
+$ composer all
+```
+
+to fix and check for coding style issues, and run the tests.
 
 If you change structure, paths, namespaces, etc., make sure you run the [autoload generator](https://getcomposer.org/doc/03-cli.md#dump-autoload):
 ```sh
-composer dump-autoload
+$ composer dump-autoload
 ```
 
 ## Contributing


simonengelhardt:
Thanks for another nice PR @localheinz. We're a bit hesitant to go down the `make` route, since we'd like to keep tool dependencies to a minimum (considering support for people on Windows or Mac OS without Xcode tools).

How about using [Composer custom commands](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands) instead (adding a `test` and a `cs` command)? I realize that wouldn't let you with no arguments (combining test and cs), but I think that would be an ok trade-off.

localheinz:
@simonengelhardt 

👍 Updated!

composer.json
```diff
@@ -28,5 +28,17 @@
         "psr-4": {
             "Facebook\\InstantArticles\\": "tests/Facebook/InstantArticles/"
         }
+    },
+    "scripts": {
+        "all": [
+            "composer cs",
```
simonengelhardt: We can simply use `@cs` here (see https://getcomposer.org/doc/articles/scripts.md#referencing-scripts)

composer.json
```diff
@@ -28,5 +28,17 @@
...
+            "composer test"
```
simonengelhardt: ^ `@test`

composer.json
```diff
@@ -28,5 +28,17 @@
...
+        ],
+        "cs": [
+            "vendor/bin/phpcbf --standard=phpcs.xml -p || vendor/bin/phpcs --standard=phpcs.xml -p"
```
simonengelhardt: We can take advantage of the PATH magic mentioned [here](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands) and simply use:

```
phpcbf --standard=phpcs.xml -p || phpcs --standard=phpcs.xml -p
```

composer.json
```diff
@@ -28,5 +28,17 @@
...
+        ],
+        "test": [
+            "vendor/bin/phpunit"
```
simonengelhardt: ^ `phpunit`

composer.json
```diff
@@ -28,5 +28,17 @@
...
+    },
+    "scripts": {
+        "all": [
+            "composer cs",
+            "composer test"
+        ],
+        "cs": [
+            "vendor/bin/phpcbf --standard=phpcs.xml -p || vendor/bin/phpcs --standard=phpcs.xml -p"
+        ],
+        "test": [
```
simonengelhardt: We could "depend" on install by adding `"@composer install",` as in the example [here](https://getcomposer.org/doc/articles/scripts.md#calling-composer-commands)

localheinz:
@simonengelhardt 

Thank you for the heads-up, wasn't aware of it!

One thing, though, adjusting `composer.json` to

```json
{
    "scripts": {
        "all": [
            "@cs",
            "@test"
        ],
        "cs": [
            "@composer install",
            "phpcbf --standard=phpcs.xml -p || phpcs --standard=phpcs.xml -p"
        ],
        "test": [
            "@composer install",
            "phpunit"
        ]
    }
}
```

and then running

```
$ rm -rf vendor && composer test
```

or 

```
$ rm -rf vendor && composer test
```

fails, because `composer install` is never executed.

Not sure why, though, composer is up-to-date:

```
$ composer --version
Composer version 1.0.3 2016-04-29 16:30:15
```

simonengelhardt:
@localheinz

Thanks for implementing all my small comments 👍

I just verified that the `@composer` syntax indeed seems to be broken. Let's assume people install Composer globally (which is what we instruct people to do in the README) and have `composer` available as an executable command for now so we can merge in what you have come up with here, file an issue on the Composer repo to fix, and then we can update to use `@composer` afterwards.

localheinz:
Thank you, @simonengelhardt!

PR #97, opened 2016-05-11
Enhancement: Allow to specify article as string

### LINKED ISSUE #96: Remove InstantArticle dependency from the Client
https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Client/Client.php#L89

`Client` has hard dependency on `InstantArticle`. The only method being used is `render` which returns a markup. 

Maybe we could do something like this to make `Client` usable outside of the SDK:
1. Check against `InstantArticleInterface` that will have `render` method
2. pass in `string` instead of `InstantArticle` - breaking change

Have you considered moving `Client` to a separate package?

Let me know what you think, I'll be happy to work on removing this dependency.

Thanks!

localheinz:
This PR

* [x] adjusts the `Client` to accept both an `InstantArticle` as well as a `string` when importing an article

Fixes #96.

### DIFF
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -81,20 +81,26 @@ class Client
     /**
      * Import an article into your Instant Articles library.
      *
-     * @param InstantArticle $article The article to import
+     * @param string|InstantArticle $article The article to import, either as a string or an InstantArticle object.
      * @param bool|false $takeLive Specifies if this article should be taken live or not. Optional. Default: false.
      */
     public function importArticle($article, $takeLive = false)
     {
-        Type::enforce($article, InstantArticle::getClassName());
+        Type::enforce($article, [
+            InstantArticle::getClassName(),
+            Type::STRING
+        ]);
+
         Type::enforce($takeLive, Type::BOOLEAN);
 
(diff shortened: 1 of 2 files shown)


src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -81,20 +81,26 @@ public static function create($appID, $appSecret, $accessToken, $pageID, $develo
     /**
      * Import an article into your Instant Articles library.
      *
-     * @param InstantArticle $article The article to import
+     * @param string|InstantArticle $article The article to import, either as a string or an InstantArticle object.
      * @param bool|false $takeLive Specifies if this article should be taken live or not. Optional. Default: false.
      */
     public function importArticle($article, $takeLive = false)
```
localheinz: @simonengelhardt @alucic

Since `$article` isn't type-hinted, this shouldn't even be a breaking change, right?

everton-rosario:
@localheinz SDK was designed to have a well formed structure of the InstantArticle markup.
There should be a Type::enforce($article, InstantArticle::getClassName()) call inside this function, in a way we don't even consider receiving string.

localheinz:
@everton-rosario 

Closing in favour of #102.

PR #100, opened 2016-05-11
Width and Height for SocialEmbed

gemedet:
This PR adds Width and Height properties to the SocialEmbed element, allowing for markup such as:
```html
<figure class="op-social">
    <iframe src="http://foo.com" width="640" height="480"></iframe>
</figure>
```


everton-rosario:
Hi @gemedet thanks for this improving code.

Are you using this widht x height on social embed?
By our reference docs this still not documented: https://developers.facebook.com/docs/instant-articles/reference/social

gemedet:
@everton-rosario It appears to be undocumented, but submitting an article with a social embed sans width and height results in:

> HTML Embeds Require Valid Width: HTML Embeds require a valid integer width in order to be displayed properly.
> HTML Embeds Require Valid Height: HTML Embeds require a valid integer height in order to be displayed properly.

everton-rosario:
Fair enough.
Im filing a task to fix the documentation there! 
Thanks @gemedet

PR #104, opened 2016-05-12
Fix: Configure whitelist filter for coverage generation

localheinz:
This PR

* [x] configures a whitelist filter for generating code coverage

💁 Without the filter, `phpunit` complains:

```
$ vendor/bin/phpunit --coverage-text
PHPUnit 4.8.24 by Sebastian Bergmann and contributors.
Warning:	No whitelist configured for code coverage

...............................................................  63 / 216 ( 29%)
............................................................... 126 / 216 ( 58%)
............................................................... 189 / 216 ( 87%)
...........................

Time: 9.07 seconds, Memory: 26.75Mb

OK (216 tests, 248 assertions)
```

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -10,4 +10,9 @@
             <directory>./tests</directory>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist>
+            <directory>./src</directory>
+        </whitelist>
+    </filter>
 </phpunit>


localheinz:
@everton-rosario @simonengelhardt 

What do you think about using http://codeclimate.com, for example? It's quite useful for tracking coverage, as well as other issues related to the code.

simonengelhardt:
@localheinz I think that's a great idea. I had considered setting up scrutinizer (since that's used in the main Facebook PHP SDK), but Code Climate seems like an even better option. I'll go ahead and set it up 👍

localheinz:
@simonengelhardt 

Nice!

Maybe this helps: https://github.com/refinery29/php-cs-fixer-config/blob/master/.travis.yml#L37-L42

everton-rosario:
@simonengelhardt @localheinz 
Is this PR going to be used?

localheinz:
@everton-rosario 

I think it only needs to be merged, right?

PR #105, opened 2016-05-14
Fix: Reference Exception from root namespace

localheinz:
This PR

* [x] asserts that `Rule::create()` and `Rule::createFrom()` throw exceptions
* [x] references `Exception` from root namespace
* [x] fixes exception messages


everton-rosario:
Thanks @localheinz

PR #107, opened 2016-05-14
Fix: Access to instance member

localheinz:
This PR

* [x] fixes an attempt to invoke the `getMessage()` method on an exception


src/Facebook/InstantArticles/Client/Helper.php
```diff
@@ -90,7 +90,7 @@ public function getPagesAndTokens($accessToken)
         try {
             $response = $this->facebook->get('/me/accounts?fields=name,id,access_token,supports_instant_articles');
         } catch (Facebook\Exceptions\FacebookResponseException $e) {
-            throw new FacebookSDKException('Graph API returned an error: ' . $e.getMessage());
+            throw new FacebookSDKException('Graph API returned an error: ' . $e->getMessage());
```
localheinz: :trollface:

PR #108, opened 2016-05-14
Fix: Remove unused import

localheinz:
This PR

* [x] removes an unused import 

@everton-rosario @simonengelhardt Would you consider switching to `fabpot/php-cs-fixer`? While `squizlabs/php_codesniffer` is a great tool, `fabpot/php-cs-fixer` has a lot more interesting fixes, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/master#usage.


everton-rosario:
Thanks @localheinz

PR #109, opened 2016-05-14
Fix: Flatten directory structure

localheinz:
This PR

* [x] flattens the directory structure

💁 We're using PSR-4, after all, why not make use of the advantages it provides over PSR-0?


everton-rosario:
This makes sense.
Also this 2 folder structure didn't created any problem.

I vote for not merging this so we can keep some standard and same structure as main PHP FB SDK (that also uses this folder structure).

everton-rosario:
Not merging this now. Ill close this PR and we can reopen or change it again in the near future.

PR #119, opened 2016-05-20
Fix: Flag take_live has been deprecated

localheinz:
This PR

* [x] uses `published` instead of `take_live`, as the latter has been deprecated

💁 For reference, see https://developers.facebook.com/docs/graph-api/reference/page/instant_articles.


everton-rosario:
@localheinz Thanks for this!
Can you rebase it, so we have it merged?

localheinz:
@everton-rosario 

👍 Rebased!

Issue #92, opened 2016-05-05
Transformer drops images from li


katazina:
The content:

```
<ul>
<li>
some text <a href="http://example.com/"><img src="http://example.com/image.jpg" width="600" height="360" /></a>
</li>
</ul>
```
After the transformer is done, there is no image.

# Expected Result

```
<ul>
        <li>some text</li>
</ul>
<figure>
        <img src="http://example.com/image.jpg"/>
</figure>
```

# Actual Result

```
<ul>
        <li>some text</li>
</ul>
```
The image is missing.

everton-rosario:
@katazina Can you help us validating this PR that fixes your issue?

katazina:
no luck with that either, like in [https://github.com/facebook/facebook-instant-articles-sdk-php/issues/91](url)

PR #124, opened 2016-06-10
Handles <img> tag inside <p> and other tags

### LINKED ISSUE #91: Transformer drops content in pharagraph if there is image too
The content:
`<p><a href="http://example.com/image.jpg">imagealt</a>Some text comes here and some more and more ... and some.</p>`

After the closing `<a>` tag, there is some text, and the sdk drops that text.

Same happens when I reorder the inside:
`<p>Some text comes here and some more and more ... and some.<a href="http://example.com/image.jpg">imagealt</a></p>`

Or if there is text before and after:
`<p>Some text comes here and some more and more ... and some.<a href="http://example.com/image.jpg">imagealt</a>Some text comes here and some more and more ... and some.</p>`

The below expexted result is just for the first content example.

# Expected Result

```
<figure>
        <img src="http://example.com/image.jpg"/>
</figure>
<p>Some text comes here and some more and more ... and some.</p>
```

# Actual Result

```
<figure>
        <img src="http://example.com/image.jpg"/>
</figure>
```

everton-rosario:
This PR

* [x] Created new Rule ImageRuleParagraph
* [x] Changed Transformer to comply with access to initial context
* [x] Added test cases to comply with this new feature 

Fixes #91 and #92


src/Facebook/InstantArticles/Transformer/Rules/ImageRuleParagraph.php
```diff
@@ -0,0 +1,87 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+namespace Facebook\InstantArticles\Transformer\Rules;
+
+use Facebook\InstantArticles\Elements\Image;
+use Facebook\InstantArticles\Elements\TextContainer;
+use Facebook\InstantArticles\Elements\InstantArticle;
+use Facebook\InstantArticles\Transformer\Warnings\InvalidSelector;
+
+class ImageRuleParagraph extends ConfigurationSelectorRule
```
diegoquinteiro: I think the naming standard is that all rule classes ends with Rule. What about `ImageInsideParagraphRule`?

src/Facebook/InstantArticles/Transformer/Transformer.php
```diff
@@ -42,6 +42,11 @@ class Transformer
     private static $allClassTypes = [];
 
     /**
+     * @var InstantArticle the initial context.
+     */
+    private $initialContext;
```
diegoquinteiro: Call it `$instantArticle`. Not every transformation needs to start with an InstantArticles, so it is not correct to assume this is the initial context.

What happens if one starts the transformation on a Paragraph instead? Nothing on this API actually enforces you to start you transformation on a InstantArticle.

In this case, this variable will be null, which is fine if we call it `$instantArticle` instead.

src/Facebook/InstantArticles/Transformer/Rules/ImageRuleParagraph.php
```diff
@@ -0,0 +1,87 @@
...
+        $image = Image::create();
+
+        // Builds the image
+        $url = $this->getProperty(self::PROPERTY_IMAGE_URL, $node);
+        if ($url) {
+            $image->withURL($url);
+            $instant_article = $transformer->getInitialContext();
+            if ($instant_article) {
+                $instant_article->addChild($image);
+            }
```
diegoquinteiro:
```php
else {
    $transformer->addWarning(
        // This new error message should be something like:
        // Could not transform Image, as no root InstantArticle was provided.
        new NoRootInstantArticleFound( ... );
    );
}

File: tests/Facebook/InstantArticles/Transformer/CMS/wp-ia.xml [hunk 4]
```diff
@@ -17,6 +17,20 @@
       </header>
       <p>Yes, peace is good for everybody!<br/>
 Man kind.</p>
+      <p>Some text comes here and some more and more.<a href="http://example.com/image1.jpg">...</a></p>
+      <figure>
+        <img src="http://example.com/image1.jpg"/>
+      </figure>
+      <p>Some text comes here and some more and more.<a href="http://example.com/image2.jpg">...</a> and some.</p>
```
diegoquinteiro: That doesn't look like the expected behavior. If the image is in the middle of the text we should close the Paragraph, add the image and open another one.

What if the image is the first thing in the Paragraph? Doesn't it make more sense to add it before the Paragraph instead of after?

This is hard as there is no concept of `appendBefore` on the SDK, and by the time you run this rule, you had already added the Paragraph to root.

My take on this would be:

1/ Do not do this for any TextContainer, but only root Paragraphs, so we know how to close/open them.
2/ The Rule should return a newly created (and appended) Paragraph as a context for the Transformer to continue processing the document.

src/Facebook/InstantArticles/Transformer/Rules/ImageInsideTextContainerRule.php
```diff
@@ -0,0 +1,98 @@
...
+        $image = Image::create();
+
+        // Builds the image
+        $url = $this->getProperty(self::PROPERTY_IMAGE_URL, $node);
+        if ($url) {
+            $image->withURL($url);
+            $instant_article = $transformer->getInstantArticle();
+            if ($instant_article) {
+                $instant_article->addChild($image);
+                //$context->enableIgnoreEmpty();
```
diegoquinteiro: 🎅

src/Facebook/InstantArticles/Elements/Element.php
```diff
@@ -86,4 +88,30 @@ protected function emptyElement($document)
         $fragment->appendChild($document->createTextNode(''));
         return $fragment;
     }
+
+    /**
+     * Method that returns the status of ignoring empty for this Element.
```
diegoquinteiro: Please comment what this actually do: "Suppress the empty tag warning on validation".

src/Facebook/InstantArticles/Elements/Element.php
```diff
@@ -86,4 +88,30 @@ protected function emptyElement($document)
...
+     */
+    public function isEmptyIgnored()
+    {
+        return $this->ignore_empty;
+    }
+
+    /**
+     * Marks this Paragraph to be ignored on isValid if it is empty.
+     */
+    public function enableIgnoreEmpty()
```
diegoquinteiro: I think it is more clear to simply use `setIgnoreEmpty(true/false);`, the words "enable" and "ignore" in the same method name are confusing when reading.

PR #126, opened 2016-06-10
Fix embedded HTML rendering

### LINKED ISSUE #62: CDATA tag being added after <script> tags
We have custom embed code for some of our content that uses javascript.  When outputing the article, the code within the `<script>` tags is wrapped with a CDATA tag.

This is the outcome:
`<figure class="op-social"><iframe><script><![CDATA[ // javascript content here ]]></script></iframe></figure>
`
What we need:
`<figure class="op-social"><iframe><script>// javascript content here</script></iframe></figure>`

This is being caused by the render() return statement in Element.php.  If we use a saveXML, it creates this CDATA tag after `<script>`, but if we use a saveHTML, there is no CDATA added.

```
return $doctype.$document->saveXML($element); // adds CDATA
return $doctype.$document->saveHTML($element); // correct (no CDATA)
```

simonengelhardt:
This PR

* [x] Refactors elements with embedded unescaped HTML
* [x] Allows passing in any unescaped HTML as string, which will eventually be passed on to the embedded browser in the Instant Articles webviews on the client.
* [x] Works around rendering CDATA sections with `saveXML` (to output valid HTML)

Fixes #62


src/Facebook/InstantArticles/Elements/Element.php
```diff
@@ -37,21 +37,11 @@ public function render($doctype = '', $formatted = false)
         $document->appendChild($element);
         $rendered = $doctype.$document->saveXML($element);
 
-        return $rendered;
-    }
+        // We can't currently use DOMDocument::saveHTML, because it doesn't produce proper HTML5 markup, so we have to strip CDATA enclosures
+        // TODO Consider replacing this workaround with a parent class for elements that will be rendered and in this class use the `srcdoc` attribute to output the (escaped) markup
+        $rendered = preg_replace('/<!\[CDATA\[(.*?)\]\]>/', '$1', $rendered);
```
anbarasan1985: I tried this and not working but `$rendered = preg_replace('/<!\[CDATA\[(.*?)\]\]>/is', '$1', $str);` works for me.

PR #128, opened 2016-06-13
Updated interactive to better align with documentation update.

dodozhang21:
From https://developers.facebook.com/docs/instant-articles/reference/embeds

```If you implement your embed using a src attribute on the <iframe> element, we recommend that you specify both the width and height of your <iframe>. Otherwise, your embedded media may be scaled incorrectly and get cropped. If the size you specify is too wide for the phone’s screen, we will automatically scale the embed down while maintaining its aspect ratio.```

Renamed the existing width property on interactive to be margin.  Added width property to support the actual width for the iframe in interactive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ composer.phar
 /vendor/
 *.ignored
 **/IgnoredTest.php
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ composer.phar
 /vendor/
 *.ignored
 **/IgnoredTest.php
-.idea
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by covenantd** [comment]
Adding `.idea` to gitignore might not be desired?

File: .gitignore [hunk 1]
```diff
@@ -2,3 +2,4 @@ composer.phar
 /vendor/
 *.ignored
 **/IgnoredTest.php
+.idea
```
localheinz: @dodozhang21 

As @covenantd pointed out, `.gitignore` should not be polluted with rules for IDEs - how about ignoring it globally?

For reference, see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore.

dodozhang21: Reverted change.

everton-rosario:
Thanks for your help @dodozhang21, @covenantd and @localheinz

Check if you'd like to join us on this event meeting we will do about IA:
http://goo.gl/forms/rZc8LKltI04bsUoF2

PR #134, opened 2016-07-01
Move validation outside of the transfomation loop

diegoquinteiro:
This PR:

- Fixes the message of the warning coming from the Validator to no include a string representation of the object (the consumer can still get a representation on their own by calling getElement).

- Removes the automatic validation from the transformation loop as there is no guarantee that the transformation will be called once and we the warnings thrown are not related to the transformation.


everton-rosario:
🚢 
Ship it.

PR #136, opened 2016-07-02
Check array before use

philipjohn:
In some circumstances, `$results` is not an array which leads to PHP "invalid argument passed to foreach" errors. This fixes those by checking `$results` is in fact an array first.

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
+++ b/src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
@@ -123,7 +123,7 @@ abstract class ConfigurationSelectorRule extends Rule
 
         $results = $domXPath->query($xpath);
 
-        if ( is_array( $results ) ) {
+        if (false !== $results) {
             foreach ($results as $result) {
                 if ($result === $node) {
                     return true;
--- a/src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
+++ b/src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
@@ -123,11 +123,14 @@ abstract class ConfigurationSelectorRule extends Rule
 
         $results = $domXPath->query($xpath);
 
-        foreach ($results as $result) {
-            if ($result === $node) {
-                return true;
+        if ( is_array( $results ) ) {
+            foreach ($results as $result) {
+                if ($result === $node) {
+                    return true;
+                }
             }
         }
+        
         return false;
     }
 


facebook-github-bot-8:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-3:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
```diff
@@ -123,11 +123,14 @@ public function matchesNode($node)
 
         $results = $domXPath->query($xpath);
 
-        foreach ($results as $result) {
-            if ($result === $node) {
-                return true;
+        if ( is_array( $results ) ) {
```
everton-rosario: @philipjohn This won't work because the method DOMXPath::query doesn't return an array. 
It returns an DOMNodeList (http://php.net/manual/en/class.domnodelist.php) which is just an Traversable with iterator.

everton-rosario:
Hey @philipjohn check my inline notes. 
I guess this PR can't be merged.

Do you have the issue that this PR is trying to fix?
Maybe we should do something else.

philipjohn:
Sorry, I should have given more detail! We're seeing this error:

> Warning: Invalid argument supplied for foreach() in /wp-content/themes/vip/plugins/facebook-instant-articles-2.11/vendor/facebook/facebook-instant-articles-sdk-php/src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php on line 126

I made this change to our deployed copy of the SDK and these errors did stop (see [this lovely graph](https://cloudup.com/cXpPPRobpzY)), so even though it's a DOMNodeList it seems to get interpreted as an array.

everton-rosario:
@philipjohn I need to double check this, because with this change, it breaks the transformer.
I'll let you know.

philipjohn:
Thanks for the code example @everton-rosario ! I made that change with the slight change that the conditional is flipped.

Are we worried about the failing build?

philipjohn:
> Are we worried about the failing build?

Ha, not now anyway :)

everton-rosario:
Yep, we have the tests running and it really helps us to figure out the issues about changing things we don't expect to be changed.
Considering this change, it was really breaking few people. 

Now it is awesome, thanks for taking the time to work on this PR @philipjohn .

Shipping now.

PR #149, opened 2016-07-18
Validation iframe width height

everton-rosario:
This PR

* [x] Adds validation to width and height in Interactive element without HTML content.
* [x] Updates test cases to comply with width and height and also adds few more test cases to cover the new feature 
* [x] Removed the SocialEmbed test, because it was deprecated.

Fixes https://github.com/Automattic/facebook-instant-articles-wp/issues/256.


src/Facebook/InstantArticles/Transformer/Warnings/validator_warning_messages.ini
```diff
@@ -24,7 +24,7 @@ H3 = "This is a text container that must not be empty nor contain only whitespac
 Header = "Header must have at least a title and/or Ad."
 Image = "Image must have a URL."
 InstantArticle = "InstantArticle relies on InstantArticleValidator."
-Interactive = "Interactive must have a src and/or HTML."
+Interactive = "Interactive embeds must have valid positive integer width and height values for an empty HTML content. In Interactive elements with HTML content, the width/height properties are optional."
```
diegoquinteiro: We should stop matching classes to messages and, instead, create context-aware messages on the validation code.

Let's create a new task for it, though, as it falls outside of the scope for this PR.

diegoquinteiro:
:shipit:

everton-rosario:
Task created here: https://github.com/facebook/facebook-instant-articles-sdk-php/issues/152

PR #150, opened 2016-07-20
Compatibility with Jetpack plugin rules

everton-rosario:
This PR
Will have the /Compat rules moved from it when WordPress unit tests are ok. While this is not true, this will be landed here.

* [x] Created compat layer to the plugin Jetpack (slideshow, tiled gallery, video)
* [x] Adds test unit cases for the new cases
* [x] Creates the configuration rules.

Related to https://github.com/Automattic/facebook-instant-articles-wp/pull/374.
Fixes #.


diegoquinteiro:
that's great work, @everton-rosario !

:shipit: and let's move this to the other repo once we get the tests figured out

everton-rosario:
Task created here: https://github.com/facebook/facebook-instant-articles-sdk-php/issues/152

PR #151, opened 2016-07-21
Add function to prepend child element to $this->children.

aminnelson:
Add the ability to prepend child elements to $this->children.


everton-rosario:
Thanks for your contribution.

Please consider joining our group @ Facebook: https://web.facebook.com/groups/629655867187397/
There we have releases, announcements, events and all exchange between all Open Source tools regarding Instant Articles.

PR #158, opened 2016-08-11
Claim URL + Submit for review

diegoquinteiro:
This PR adds two new methods to the client, supporting 2 new APIs we'll be releasing:

- [x] Review submission API
- [x] Domain claiming API

### DIFF
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -250,7 +250,7 @@ class Client
      */
     public function claimURL($url)
     {
-        $response = $this->facebook->post('me/claimed_urls?url=' + $url);
+        $response = $this->facebook->post($this->pageID . '/claimed_urls?url=' . $url);
         $node = $response->getGraphNode();
         $error = $node->getField('error');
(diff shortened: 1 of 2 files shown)


src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -242,4 +242,42 @@ public function getArticlesURLs($limit = 10, $development_mode = false)
 
         return $articleURLs;
     }
+
+    /**
+     * Claims an URL for the page
+     *
+     * @param string $url The root URL of the site without the protocol (ex: exemple.com/path)
+     */
+    public function claimURL($url)
```
everton-rosario: What if informing $url with protocols?
Shouldn't be treated/rejected in code?

everton-rosario:
Add a test case and treatment when invalid $url is informed.
Like:
* http://example.com/path
* ftp://example.com/path
* invalidcontent/path

diegoquinteiro:
@everton-rosario I've fixed the http:// and https:// cases;

The ftp:// and invalidcontent cases will be rejected by the server.

everton-rosario:
:shipit:

PR #160, opened 2016-08-19
Encapsulate encoding stuff on Transformer.

yssk22:
This PR

* [x]  adds the function to encapsulate multibyte encoding transformation. 

We are seeing bunch of issues related with content encoding of `DOMDocument->loadHTML()` in Wordpress plugin so it is worth for SDK to wrap this part and any developers can pass string to process transformation.


yssk22:
will PR again once issues on travis is fixed.

everton-rosario:
Nice catch @yssk22.
Looking forward to your PR with this cases covered.

PR #161, opened 2016-08-19
Introduce transformString() function to encapsulate the transformation process in multibyte environment.

yssk22:
This PR

* [x] introduce a new transform function that supports multibyte encoding environment.

Current `transform` function requires `DOMDocument` which requires developers to  support multibyte encoding on their side. There are several ways to do that and this issue is one of the common issues in PHP environment. 

The new function gets rid of this effort from developers and they can just pass HTML string to get the transformed HTML.

### DIFF
--- a/src/Facebook/InstantArticles/Elements/GeoTag.php
+++ b/src/Facebook/InstantArticles/Elements/GeoTag.php
@@ -101,7 +101,10 @@ class GeoTag extends Element
 
         // Required script field
         if ($this->script) {
-            $element->appendChild($document->createTextNode($this->script));
+            // script may contain html entities so import it as CDATA
+            $element->appendChild(
+                $element->ownerDocument->importNode(new \DOMCdataSection($this->script), true)
+            );
         }
 
         return $element;
--- a/src/Facebook/InstantArticles/Transformer/Transformer.php
+++ b/src/Facebook/InstantArticles/Transformer/Transformer.php
@@ -122,6 +122,36 @@ class Transformer
 
     /**
      * @param InstantArticle $context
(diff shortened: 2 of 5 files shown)


everton-rosario:
@yssk22 This looks great!


Can you add a test case that uses CDATA escaping content for XHTML compatibility? 

Like this:
```
      <figure class="op-interactive">
         <iframe src="http://example.com/custom-interactive.xhtml" class="column-width" height="60">
           <h1>カスタムコード</h1>
           <script>
               <![CDATA[alert('テスト');]]>
           </script>
         </iframe>
         <figcaption>このグラフィックは素晴らしい。</figcaption>
       </figure>
```

Just to be sure we won't have any problem with XHTML and double encoding of CDATA.

yssk22:
@everton-rosario thanks for reviewing PR. yes, XHTML compatibility is something we need to consider. 

Currently, `loadHTML` (`DOMDocument`) does not do anything on parsing cdata and the script content can include CDATA tags. This causes an issue on `Elements\GeoTag->isValid()` and the code omit script in CDATA even we use the original transform function.

One of the workaround for developers is to write

```
<script type="text/javascript">
// <![CDATA[
script here
// ]]>
</script>
```

And this is a typical solution for browsers who cannot understand XHTML.

On the other hand, we can remove '<![CDATA[' and ']]>' string from the original content. do you want to do this on SDK in this PR?

everton-rosario:
Thanks for raising it @yssk22.
Lets do it in another PR.
Filed one issue with that: #163

everton-rosario:
Shipping this.

🚢

PR #162, opened 2016-08-21
Add HardcodedElementGetter

diegoquinteiro:
This PR adds a new getter allowing the output to contain markup not originally in the source. These hardcoded snippets are intended to be used with the MultipleElementGetter to append or prepend code to embeds.

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Getters/FragmentGetter.php
+++ b/src/Facebook/InstantArticles/Transformer/Getters/FragmentGetter.php
@@ -12,7 +12,7 @@ use Facebook\InstantArticles\Validators\Type;
 use Facebook\InstantArticles\Transformer\Transformer;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 
-class HardcodedElementGetter extends AbstractGetter
+class FragmentGetter extends AbstractGetter
 {
     /**
      * @var string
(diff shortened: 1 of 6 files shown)


diegoquinteiro:
@everton-rosario please review =)

everton-rosario:
@diegoquinteiro This looks like a great feature!

Though, "Hardcoded" seems not a good name IMO.

> "hard coding" def: putting something fixed into your source code. Any change needs code change.

Since this is configured, I'd vote for something like:

- GlobalGetter
- ParentScopeGetter
- GlobalScopeGetter
- FreeScopeGetter
- etc

Considering the naming of the getter, Im pro for adding more test cases:

A test case where selector is a sibling inner same container element
```
<container>
    <selector>Content from selector</selector>
    <freeselector>Content from free selector</freeselector>
</container>
```

A test case where selector is within any hierarchy and the free selector is after the selector and outside same block
```
<container>
    <selector>Content from selector</selector>
</container>
<freeselector>Content from free selector</freeselector>
```

A test case where selector is inner any hierarchy and the free selector is before and outside the selector hierarchy
```
<freeselector>Content from free selector</freeselector>
<container>
    <selector>Content from selector</selector>
</container>
```

A test case where selector is outter and the free selector is within any hierarchy.
```
<selector>Content from selector</selector>
<container>
    <freeselector>Content from free selector</freeselector>
</container>
```

diegoquinteiro:
@everton-rosario I think you missed the point of this particular getter. It is **not** selecting anything globally, it is inserting arbitrary markup on the output. You do not provide a selector, but instead a markup fragment.

I'll change the name to FragmentGetter.

everton-rosario:
Makes sense @diegoquinteiro .
Im working on a PR to do what I mentioned.

Shipping this one.

Issue #171, opened 2016-10-10
Latest Laravel requires other version of Symfony's css-selector


ronaldgj:
# Steps required to reproduce the problem

1. Require version 3.* of symfony's css-selector

# Expected Result

* No errors when installing this SDK with composer in Laravel

# Actual Result

* Errors, because Laravel needs version 3.* and Facebook 2.8.

everton-rosario:
@ronaldgj Thanks for flagging this.
So far we are not aiming to move to the newest version of CSS selector because it requires PHP > 5.5.+ and we still need to have it compatible to PHP >= 5.4.

What I recommend you to do is:
* Checkout the master version of this component
* Change composer.json to set required version of the CSS Selector:
  ```
 "require": {
        "php": "^5.4 || ^7.0",
        "symfony/css-selector": "3.*",
        "facebook/graph-sdk": "~5.0",
        "apache/log4php": "2.3.0"
    }
```
* Run `composer update` then `composer install`

It will build a version compatible as you need.

Thanks for flagging this issue.

ronaldgj:
For now I changed the composer.json of Laravel back to 2.8. I saw Laravel only uses it for 'tests'. For now I'm not using that so this is a temporary solution for me. I'll will try this soon and let you know if it worked!

Thank you for your respons!

kamranata:
Hi @everton-rosario. I saw you added to your composer.json 
`"php": "^5.4 || ^5.6 || ^7",
"symfony/css-selector": "^2.8 || ^3.1 || ^4.1",`

But when I use `composer require facebook/facebook-instant-articles-sdk-php` composer uses another json file. `
        "php": "^5.4 || ^7.0",
        "symfony/css-selector": "^2.8 || ^3.0",`

Can you fix this?

everton-rosario:
So you're mentioning a conflict with the symfony/css-selector. Can you propose a PR so I can understand better what you are proposing?

Thanks for reporting it.

PR #184, opened 2016-11-22
allow ^3.0 version for symfony/css-selector

### LINKED ISSUE #45: Installation fails, requires older version of syfony/css-selector

ahilles107:
This PR allows `symfony/css-selector` in version `^3.0` - it doesn't require it - just allow if other dependencies allows it. 

* [x] allow usage of that library in symfony 3.x apps.

Fixes #45, #48, #171 .

Tests are passing when when Symfony 2.8 and 3.x is added to dependencies (and vendors are updated).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     }],
     "require": {
         "php": "^5.4 || ^7.0",
-        "symfony/css-selector": "2.8.*",
+        "symfony/css-selector": "2.8.* || ^3.0",
         "facebook/graph-sdk": "~5.0",
         "apache/log4php": "2.3.0"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "86691e0c3820750436470d6924d8b59f",
-    "content-hash": "064cf914d1becccc9b36fb304b2b632b",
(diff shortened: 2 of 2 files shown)


ahilles107:
@diegoquinteiro - what do You think?

everton-rosario:
@ahilles107 Thanks for reporting it and helping us improve quality of this plugin!

We are still using this old version of symfony just because of our compatibility with PHP 5.4.
Version 3.0 wasn't compatible with PHP 5.4. 

Is the issue you are facing related to the symfony 3.0? Just by upgrading it fixes for you?
Can you share more details about this issue?

ahilles107:
@everton-rosario i understand why you want to keep 2.8 (compatibility with php 5.4). And this change don't break it. It just allow use library in project based on SYmfony 3.x (or just other php project which use `symfony/css-selector` in version bigger than 2.8. 

Allowing both (`2.8.* || ^3.0`) versions fix issue with incompatibility with Symfony 3.x and still allow to work with library on php 5.4 (when update is called on 5.4, or one of other libabries require `symfony/css-selector` in version 2.8 then this version will be seleced by composer, if not then highest possible will be installed).

everton-rosario:
Makes sense!
Great point. Will get it through the tests to double check, but good to go.

PR #185, opened 2016-11-25
Added feature of sufix and prefix fixed for string getters

everton-rosario:
This PR

* [x] Adds suffix and prefix to string getters


src/Facebook/InstantArticles/Transformer/Getters/StringGetter.php
```diff
@@ -40,16 +56,52 @@ public function withAttribute($attribute)
         return $this;
     }
 
+    /**
+     * @param string $prefix
+     *
+     * @return $this
+     */
+
```
diegoquinteiro: 💅  remove empty line between php-doc and method declaration

PR #190, opened 2017-01-03
Issue 182 force rescrape

### LINKED ISSUE #182: Force OG Re-scrape on post update
Follow best practice and force a re-scrape of the OG tags whenever a post is updated.

https://developers.facebook.com/docs/sharing/opengraph/using-objects#update

jppgks:
Closes #182.

### DIFF
--- a/src/Facebook/InstantArticles/Client/Client.php
+++ b/src/Facebook/InstantArticles/Client/Client.php
@@ -88,7 +88,7 @@ class Client
      *
      * @return int The submission status ID. It is not the article ID. (Since 1.3.0)
      */
-    public function importArticle($article, $published = false)
+    public function importArticle($article, $published = false, $forceRescrape = false)
     {
         Type::enforce($article, 'Facebook\InstantArticles\Elements\InstantArticleInterface');
         Type::enforce($published, Type::BOOLEAN);
@@ -103,8 +103,10 @@ class Client
           'development_mode' => $this->developmentMode,
         ]);
 
-        // Re-scrape Graph object for article URL
-        $this->scrapeArticleURL($article->getCanonicalURL());
+        if ($forceRescrape) {
+            // Re-scrape Graph object for article URL
+            $this->scrapeArticleURL($article->getCanonicalURL());
+        }
 
         return $response->getGraphNode()->getField('id');
     }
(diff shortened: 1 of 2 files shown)


jppgks:
@everton-rosario I tested it with WordPress `Draft` to `Published` status. Once the post gets published, the URL is re-scraped and the right information shows in the [Sharing Debugger](https://developers.facebook.com/tools/debug/sharing/).

jppgks:
@everton-rosario If we don't want the URL to be scraped before the page is accessible, we could check the HTTP status code before scraping?

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -103,10 +103,26 @@ public function importArticle($article, $published = false)
           'development_mode' => $this->developmentMode,
         ]);
 
+        // Re-scrape Graph object for article URL
```
everton-rosario: Consider checking a parameter (add a new one on the importArticle method) that can be default to false.
This parameter could be: $forceRescrape defaults to false.
it would be:
```
   if ($forceRescrape) {
      // do the rescrape
   } 
```

everton-rosario:
@jppgks I guess it is ok: https://github.com/facebook/facebook-instant-articles-sdk-php/pull/190#issuecomment-270100718

One thing to consider is to not enforce this as a required feature. I'd add that parameter to make this an optional rescrape.

src/Facebook/InstantArticles/Client/Client.php
jppgks: @everton-rosario Done!

jppgks: @everton-rosario Done. Also updated tests to reflect this behaviour.

jppgks:
🎉

jppgks:
@everton-rosario Quick question: are we waiting on other PRs before this can be merged?

everton-rosario:
@jppgks Merged!
Im not sure if you are already in, but you can join a FB group:
https://web.facebook.com/groups/629655867187397/

There we have releases, announcements, events and all exchange between all Open Source tools regarding Instant Articles.

demoive:
Just had a thought about this. The re-scrape endpoint (`POST http://graph.facebook.com/?id={url}&scrape=true`) works regardless of an access token. The current implementation of `scrapeArticleURL()` within `Client` means we're restricting ourselves to it being initialized, right? Would it be worth changing this so that this rescrape functionality could be triggered without a Facebook App?

cc @everton-rosario

Issue #191, opened 2017-01-06
Can we have the ability to remove a child from an InstantArticle class object?


despinoUY:
As @dashaluna asked in other Facebook project (https://github.com/Automattic/facebook-instant-articles-wp/issues/529), is there any chance to have a function to remove a child from an InstantArticle object? If it's not, a function to load a completely different array of Element objects as Children would be great!

Greets!

everton-rosario:
This is fixed and will be on release v1.5.5

PR #192, opened 2017-01-08
Issue 191 solution?

despinoUY:
I believe it solves https://github.com/facebook/facebook-instant-articles-sdk-php/issues/191.

### DIFF
--- a/src/Facebook/InstantArticles/Elements/InstantArticle.php
+++ b/src/Facebook/InstantArticles/Elements/InstantArticle.php
@@ -225,7 +225,7 @@ class InstantArticle extends Element implements Container, InstantArticleInterfa
      *
      * @return $this
      */
-    public fucntion withChildren($children)
+    public function withChildren($children)
     {
         Type::enforceArrayOf(
             $children,
@@ -262,7 +262,7 @@ class InstantArticle extends Element implements Container, InstantArticleInterfa
      *
      * @return $this
      */
-    public fucntion deleteChild($children)
+    public function deleteChild($children)
     {
         Type::enforce($index, Type::INTEGER);
         $children = NULL;
@@ -285,7 +285,8 @@ class InstantArticle extends Element implements Container, InstantArticleInterfa
      */
     public function addChild($child)
     {
-        Type::enforce($child,             [
+        Type::enforce($child,
+             [
                 Ad::getClassName(),
                 Analytics::getClassName(),
                 AnimatedGIF::getClassName(),
--- a/src/Facebook/InstantArticles/Elements/InstantArticle.php
+++ b/src/Facebook/InstantArticles/Elements/InstantArticle.php
@@ -262,7 +262,7 @@ class InstantArticle extends Element implements Container, InstantArticleInterfa
      *
      * @return $this
      */
-    public function deleteChild($children)
+    public function deleteChild($index)
     {
         Type::enforce($index, Type::INTEGER);
         $children = null;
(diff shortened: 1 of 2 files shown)


src/Facebook/InstantArticles/Elements/InstantArticle.php
```diff
@@ -219,6 +219,64 @@ public function withFooter($footer)
...
+        );
+        $this->chidren = $children;
+
+        return $this;
+    }
+
+    /**
+     * Replace all the children within this InstantArticle
+     *
+     * @param Type::INTEGER $index The index of the element to delete
```
everton-rosario: Looks like this parameter is missing/changed. Can you double check?

src/Facebook/InstantArticles/Elements/InstantArticle.php
```diff
@@ -219,6 +219,64 @@ public function withFooter($footer)
...
+     *                             in the array of children.
+     *
+     * @return $this
+     */
+    public function deleteChild($children)
```
everton-rosario: Im not sure if this should be $index or if your logic should be something else.
Can you double check?

src/Facebook/InstantArticles/Elements/InstantArticle.php
```diff
@@ -219,6 +219,64 @@ public function withFooter($footer)
...
+    {
+        Type::enforce($index, Type::INTEGER);
+        $children = null;
+        foreach ($this->children as $childIndex => $child) {
```
everton-rosario: This loop looks broken because of previous comments. 
Please take a look.

src/Facebook/InstantArticles/Elements/InstantArticle.php
despinoUY: Fixed! Sorry for that, the description was OK but the param name in the implementation was wrong. :(

despinoUY:
Finally fixed! :D

despinoUY:
@everton-rosario Have you checked this?

oussamaesprit:
How to run code and use

src/Facebook/InstantArticles/Elements/InstantArticle.php
```diff
@@ -219,6 +219,64 @@ public function withFooter($footer)
...
+     *
+     * @param Type::INTEGER $index The index of the element to be deleted
+     *                             in the array of children.
+     *
+     * @return $this
+     */
+    public function deleteChild($index)
+    {
+        Type::enforce($index, Type::INTEGER);
+        $children = null;
```
everton-rosario: $children should never be null.

You gotta initialize it with:
`$children = [];`

See the class initialization. If the index informed is the only child of the array, this will assign a null array to the `$this->children` leading to a broken status of this element.

tests/Facebook/InstantArticles/Elements/InstantArticleTest.php
```diff
@@ -473,4 +473,133 @@ public function testIsRTLEnabled()
 
         $this->assertEquals($expected, $result);
     }
+
+    public function testDeleteChildren()
```
everton-rosario: Please add a testDeleteOnlyChild() test case, where you add only 1 paragraph and delete it. This will throw that error on an invalid state about `$this->children` array.

This is why we need it to be initialized with an empty array, not a null one.

despinoUY:
Don't worry, @everton-rosario! I've just made those changes. Waiting for the results of the tests! :)

despinoUY:
Done @everton-rosario!

everton-rosario:
@despinoUY If you still didn't joined please join our group @ Facebook: https://web.facebook.com/groups/629655867187397/
There we have releases, announcements, events and all exchange between all Open Source tools and community regarding Instant Articles.

despinoUY:
Thanks @everton-rosario, I've already joined some days ago. Loved this community! :D

oussamaesprit:
Hi,Please accept my invitation to join the group .

2017-02-07 15:07 GMT+01:00 Danilo Espino <notifications@github.com>:

PR #199, opened 2017-03-09
Adds parameter for formatting content ingested

everton-rosario:
[x] Adds a treatment for "NULL" informed parameter as content type
[x] Implements default value as false for formatted articles
[x] Respects and overwrites if value informed

Fixes https://github.com/Automattic/facebook-instant-articles-wp/issues/625

It is a pré-requisite for: https://github.com/Automattic/facebook-instant-articles-wp/pull/635

### DIFF
--- a/src/Facebook/InstantArticles/Elements/InstantArticle.php
+++ b/src/Facebook/InstantArticles/Elements/InstantArticle.php
@@ -396,6 +396,7 @@ class InstantArticle extends Element implements Container, InstantArticleInterfa
 
     public function render($doctype = '<!doctype html>', $format = false)
     {
+        $doctype = is_null( $doctype ) ? '<!doctype html>' : $doctype;
         return parent::render($doctype, $format);
     }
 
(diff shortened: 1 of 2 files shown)


src/Facebook/InstantArticles/Elements/InstantArticle.php
```diff
@@ -396,6 +396,7 @@ public function addMetaProperty($property_name, $property_content)
 
     public function render($doctype = '<!doctype html>', $format = false)
     {
+        $doctype = is_null($doctype) ? '<!doctype html>' : $doctype;
```
sakatam: This can be a matter of personal preference, but how about changing the function signature to 
```php:
public function render($doctype = null, $format = false)
```
so you can avoid the recurring '<!doctype html>'.

everton-rosario: @sakatam This would be a breaking change if somebody is already using the default method call (using the default value) which leads to a `<!doctype html>` addition by default. What do you think about it?

sakatam: My understanding is that even with the signature change, if somebody is using this method as either `$article->render()` or `$article->render(null)`, the `$doctype` will be set to `<!doctype html>` because of the line you added. Maybe I am missing something?

everton-rosario: Thats true, and expected.
If needed to override that default parameter, you gotta inform `''` (empty string/quotes)

PR #203, opened 2017-03-24
Added accessors for style and isRTLEnabled

pestevez:
This PR

* [x]  Adds two functions to access the values of `style` and `isRTLEnabled`


src/Facebook/InstantArticles/Elements/InstantArticle.php
```diff
@@ -381,6 +389,14 @@ public function getChildren()
     }
 
     /**
+     * @return boolean if this article is Right-to-left(RTL).
+     */
+    public function getIsRTLEnabled()
```
everton-rosario: method should be just `isRTLEnabled()`.

PR #217, opened 2017-05-10
Get Author name from address text

pestevez:
This PR

* [x] Allows Author names to be parsed from the `<address>` node text (when it has no `<a>` descendant), which is also supported in IA.
* [x] Updates the selector to enforce the `<address>` to be a descendant of the `<header>` element.
* [x] Adds unit tests to parse author names with and without `<a>` nodes.

### DIFF
--- a/src/Facebook/InstantArticles/Parser/instant-articles-rules.json
+++ b/src/Facebook/InstantArticles/Parser/instant-articles-rules.json
@@ -248,6 +248,15 @@
             },
             {
                 "class": "AuthorRule",
+                "selector" : "address",
+                "properties" : {
+                    "author.name" : {
+                        "type" : "string"
+                    }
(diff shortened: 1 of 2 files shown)


src/Facebook/InstantArticles/Parser/instant-articles-rules.json
```diff
@@ -271,6 +271,15 @@
                 }
             },
             {
+                "class": "AuthorRule",
```
everton-rosario: If you put this rule above the other, it wont be necessary to use the `not(a)` into the selector.

pestevez:
Thanks, @everton-rosario, good point.

I agree, I changed the order of the rules and simplified the selector. Please let me know if this one looks good to you.

PR #218, opened 2017-05-12
Made the readme a little more explicit, changed the structure a little

vkama:
Updated the readme with more explicit information/steps

Changed the structure to become more clear to the two types of audiences looking at the SDK:
1. developers that want to use it, to integrate it into their product
2. developers that want to contribute to the sdk development

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,19 +11,31 @@ The SDK consists of three components:
 - **Client**: A simple wrapper around the [Instant Articles API](https://developers.facebook.com/docs/instant-articles/api), which can be used for publishing Instant Articles on Facebook. The client provides a CRUD interface for Instant Articles as well as a helper for authentication. The client depends on the main [Facebook SDK for PHP](https://github.com/facebook/facebook-php-sdk-v4) as an interface to the Graph API and Facebook Login.
 
 ## Quick Start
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,19 +11,31 @@ The SDK consists of three components:
 - **Client**: A simple wrapper around the [Instant Articles API](https://developers.facebook.com/docs/instant-articles/api), which can be used for publishing Instant Articles on Facebook. The client provides a CRUD interface for Instant Articles as well as a helper for authentication. The client depends on the main [Facebook SDK for PHP](https://github.com/facebook/facebook-php-sdk-v4) as an interface to the Graph API and Facebook Login.
 
 ## Quick Start
-You can find examples on how to use the different components of this SDK to integrate with your CMS in the [Getting Started section](https://developers.facebook.com/docs/instant-articles/sdk/#getting-started) of the documentation.
 
-## Installation
+The Facebook Instant Articles PHP SDK can be installed with the [Composer](https://getcomposer.org/) dependency manager by running this command on your project's root folder:
```
everton-rosario: @vkama What do you think about adding instructions to download it from packagist also?
People who are not aware of how composer works, could also use this SDK library.

vkama: From my understanding, packagist is only a meta-data website. Meaning that when we do a `composer install` the script actually goes to the package's githubs and downloads the sources

PR #220, opened 2017-05-15
Fix istextempty

vkama:
Elements that have text like Paragraph, Caption, etc that inherit from TextContainer were affected by this bug.

The problem was that TextContainer::isValid() calls Type::isTextEmpty() and that function had strip_tags($text) that was removing valid captions like "<3 this sdk". It would interpret this string as an html tag and remove the entire line.

I included <br> as a "blank character" too, following the trend of nbsp; and tweaked the regex.

Also updated some tests too.

### DIFF
--- a/src/Facebook/InstantArticles/Elements/TextContainer.php
+++ b/src/Facebook/InstantArticles/Elements/TextContainer.php
@@ -41,6 +41,15 @@ abstract class TextContainer extends Element implements Container
     }
 
     /**
+     * Clears the text.
+     *
+     */
+    public function clearText()
+    {
+        $this->textChildren = [];
+    }
+
+    /**
      * @return string[]|FormattedText[]|TextContainer[] All text token for this text container.
      */
     public function getTextChildren()
--- a/src/Facebook/InstantArticles/Elements/TextContainer.php
+++ b/src/Facebook/InstantArticles/Elements/TextContainer.php
@@ -23,7 +23,7 @@ abstract class TextContainer extends Element implements Container
     /**
      * @var array The content is a list of strings and FormattingElements
      */
-    private $textChildren = [];
+    private $textChildren = array();
 
     /**
      * Adds content to the formatted text.
@@ -42,11 +42,10 @@ abstract class TextContainer extends Element implements Container
 
     /**
      * Clears the text.
-     *
      */
     public function clearText()
     {
-        $this->textChildren = [];
+        $this->textChildren = array();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Facebook/InstantArticles/Elements/TextContainer.php
```diff
@@ -41,6 +41,15 @@ public function appendText($child)
     }
 
     /**
+     * Clears the text.
+     *
```
everton-rosario: Remove this line.

src/Facebook/InstantArticles/Elements/TextContainer.php
```diff
@@ -41,6 +41,15 @@ public function appendText($child)
...
+     */
+    public function clearText()
+    {
+        $this->textChildren = [];
```
everton-rosario: PHP version retro use:
`array()`

src/Facebook/InstantArticles/Validators/Type.php
```diff
@@ -396,9 +398,10 @@ public static function isTextEmpty($text)
             return true;
         }
         // Stripes empty spaces, &nbsp;, <br/>, new lines
-        $text = strip_tags($text);
-        $text = preg_replace("/[\r\n\s]+/", "", $text);
+        $text = preg_replace("/\s+/", "", $text);
         $text = str_replace("&nbsp;", "", $text);
+        $text = str_replace("<br>", "", $text);
+        $text = str_replace("<br/>", "", $text);
```
diegoquinteiro: Actually, let's use this opportunity to fix a problem with the use of `strip_tags` in this method. Here, if we're looking at the text content of any Element, it is *not* an HTML snippet, and we do convert it to HTML Entities before outputting. We should consider any text content as plain text.

Example:

```php
Paragraph $p1 = new Paragraph();
$p1->appendText("This is how you make line breaks in HTML:");
Paragraph $p2 = new Paragraph();
$p2->appendText("<br>");
$p1->isValid(); // Should return true.
$p2->isValid(); // Should return true.
````

Let's remove these two lines and fix tests accordingly.

File: src/Facebook/InstantArticles/Elements/TextContainer.php [hunk 4]
```diff
@@ -23,7 +23,7 @@
     /**
      * @var array The content is a list of strings and FormattingElements
      */
-    private $textChildren = [];
+    private $textChildren = array();
```
diegoquinteiro: 💅

PR #224, opened 2017-06-12
Fix: Remove accidentally checked in .DS_Store

localheinz:
This PR

* [x] removes an accidentally checked in `.DS_Store`

Follows https://github.com/facebook/facebook-instant-articles-sdk-php/pull/186.


localheinz:
Thank you, @everton-rosario!

PR #227, opened 2017-06-12
Fix: Remove return annotation

localheinz:
This PR

* [x] removes a `@return` annotation from the docblock of a constructor


localheinz:
Thank you, @everton-rosario!

PR #228, opened 2017-06-12
Fix: Use more appropriate assertion

localheinz:
This PR

* [x] uses more appropriate assertions in tests


everton-rosario:
Nice fix!

localheinz:
Thank you, @everton-rosario!

PR #244, opened 2017-08-30
Full HTML Parser

everton-rosario:
This PR

* [x] Creates a GlobalRule
* [x] Creates a Date getter
* [x] Tests for the case of parsing a full HTML to create an Instant Article


everton-rosario:
Trashed branch.

PR #246, opened 2017-09-29
Converting to HHVM

mburak:
Changed <?php to <?hh and applied changes to make the code hhvm compliant so tests run successfully.


src/Facebook/InstantArticles/Elements/Author.php
```diff
@@ -198,7 +198,9 @@ public function toDOMElement($document = null)
         $element->appendChild($ahref);
 
         // Appends author description
-        $element->appendChild($document->createTextNode($this->description));
+        if ($this->description) {
```
everton-rosario: Probably a better strategy will be to use the Type::isTextEmpty, instead of just "if" also.

src/Facebook/InstantArticles/Elements/GeoTag.php
```diff
@@ -11,7 +11,7 @@
 use Facebook\InstantArticles\Validators\Type;
 
 /**
- * Class Map
+ * Class GeoTag
```
everton-rosario: Nice catch. Worth it to send this diff to the PHP main open source  as well 👍

src/Facebook/InstantArticles/Transformer/Getters/ElementGetter.php
```diff
@@ -53,7 +53,8 @@ public function withSelector($selector)
     public function get($node)
     {
         $elements = self::findAll($node, $this->selector);
-        if (!empty($elements) && property_exists($elements, 'length') && $elements->length !== 0) {
+        // var_dump($elements instanceof DOMNodeList);
```
everton-rosario: Remove this.

src/Facebook/InstantArticles/Transformer/Getters/ElementGetter.php
```diff
@@ -53,7 +53,8 @@ public function withSelector($selector)
...
+        if (!empty($elements) /*&& property_exists($elements, 'length')*/ && $elements->length !== 0) {
```
everton-rosario: We need to be sure the element has the property "length" and we should be sure it is a number and not zero (needs to be positive number).

src/Facebook/InstantArticles/Transformer/Rules/InstantArticleRule.php
```diff
@@ -80,7 +80,7 @@ public function apply($transformer, $instant_article, $node)
             $instant_article->disableAutomaticAdPlacement();
         } else {
             $instant_article->enableAutomaticAdPlacement();
-            $pairs = explode(' ', $auto_ad_placement, 2);
+            $pairs = explode(' ', $auto_ad_placement ?: '', 2);
```
everton-rosario: Is this intentional?
`$auto_ad_placement ? $auto_ad_placement : ''`

everton-rosario:
To this same Pull request, you can change the entries about running it against few older PHP versions, since it won't be covered anymore for this branch.

It is just about removing the versions from the `.travis.yml`

src/Facebook/InstantArticles/Transformer/Getters/ElementGetter.php
mburak: The problem here is that property_exists doesn't work with HHVM DOMNodeList, it looks like it's creating the property dynamically so it always returns false. Is there anything else we can check? When this might not come with length?

src/Facebook/InstantArticles/Transformer/Rules/InstantArticleRule.php
mburak: yes, just avoiding passing null to explode as now it's typed and it doesn't allow it.

mburak:
Updated pr with requested changes. I couldn't change the name for ChildrenContainer back to Container because HHVM is still complaining even using full namespace. What would you suggest to keep retro-compatibility?

PR #247, opened 2017-10-09
removed useless folders by using better autloading

OskarStark:
No need for the folders


timjacobi:
Can you please merge master into this branch?

OskarStark:
done @timjacobi

timjacobi:
@OskarStark the checks are still failing. But actually I think the folder structure is ok like this.

OskarStark:
For the future it would be great if you decide sth like this before I do the rebase stuff...

PR #255, opened 2017-10-17
Fix slideshow image caption not displayed in case of empty caption title, but available credits

hctom:
This PR changes the `SlideshowImageRule::apply()` method to display slideshow image caption in case of empty caption title, but available credits.

This ensures that the slideshow image caption is always displayed, when either the title or the credits have some output.


hctom:
Okay... i guess you can forget about this. Unfortunately, the caption is never displayed when there is no text/title, but credits.

PR #266, opened 2017-10-23
using full path for BaseHTMLTestCase (for www compatibility)

mburak:
This PR

* Changes tests to extend BaseHTMLTestCase with full path so it doesn't have HHVM lints in www.


everton-rosario:
@mburak Make sure we have this PR landed on Master too.

PR #272, opened 2017-10-27
Fix #271

### LINKED ISSUE #271: @meta "fb:op-recirculation-ads" were removed when Transformer?
1. Prepare input html and make sure html contains `meta` tag "fb:op-recirculation-ads", like this:

```
<meta property="fb:op-recirculation-ads" content="placement_id=xxxxxxxxx">
```

2. Load rules and load input html by Transformer

```
$rules = file_get_contents(__DIR__ . "/facebook-sdk-php/vendor/facebook/facebook-instant-articles-sdk-php/src/Facebook/InstantArticles/Parser/instant-articles-rules.json", true);
// Create the transformer and loads the rules
$transformer = new Facebook\InstantArticles\Transformer\Transformer();
$transformer->loadRules($rules);

libxml_use_internal_errors(true);
$document = new \DOMDocument();
$v_news_html = mb_convert_encoding($v_news_html, 'HTML-ENTITIES', "UTF-8");
$document->loadHTML($v_news_html);
libxml_use_internal_errors(false);

$transformer->transform($instant_article, $document);
```
I added some custom rules for this `meta` tag, but anything change

```
{
    rules: [...
        ,{
                "class" : "PassThroughRule",
                "selector" : "head meta"
            } ] }
```

3. Import Instant Article

```
$__fbClient->importArticle($instant_article);
```

How to add that meta tag?

tungvn:
This PR

* [x] Fix add meta tag for recirculation ads

Fixes #271 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     - php: 5.6
     - php: 7
     - php: 7.1
-      env: WITH_CS=true
+      env: WITH_CS=true GITHUB_TOKEN=fbc9f8b635109c0c7ad64013ff8346908389f5f9
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 # Fix TravisCI, add github token
 language: php
 
+env:
+  - GITHUB_TOKEN=fbc9f8b635109c0c7ad64013ff8346908389f5f9
+
 matrix:
   include:
     - php: 5.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+# Fix TravisCI, add github token
 language: php
 
 matrix:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,13 +1,17 @@
...
+
 matrix:
   include:
     - php: 5.4
     - php: 5.5
     - php: 5.6
     - php: 7
     - php: 7.1
-      env: WITH_CS=true
+      env: WITH_CS=true GITHUB_TOKEN=fbc9f8b635109c0c7ad64013ff8346908389f5f9
```
localheinz: Do we really want to expose a GitHub API token like that?

tungvn: This is a token without any selected permissions. But I think we should hide it.

localheinz: @tungvn 

If we encrypt it, it won’t be available for pull request builds of forks, so I think we can just leave it out. 

See https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml:

>Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.
>
>If you define a variable with the same name in .travis.yml and in the Repository Settings, the one in `.travis.yml` takes precedence. If you define a variable in `.travis.yml` as both encrypted and unencrypted, the one defined later in the file takes precedence.

PR #275, opened 2017-11-02
convert &amp;{entitityname}; in output to &{entityname};

timjacobi:
All classes extending Element make use of createTextNode which converts symbols
to their respective HTML entities e.g. `&` -> `&amp;`

The inputs are not sanitized and therefore if they contain entities the & part
will be converted to &amp; e.g. `&quot;` -> `&amp;quot;`

This commit adds a test to prove this and fixes the problem by converting the
render output back.

Fixes https://github.com/Automattic/facebook-instant-articles-wp/issues/693

### DIFF
--- a/tests/Facebook/InstantArticles/Transformer/TransformerTest.php
+++ b/tests/Facebook/InstantArticles/Transformer/TransformerTest.php
@@ -11,6 +11,7 @@ namespace Facebook\InstantArticles\Transformer;
 use Facebook\InstantArticles\Elements\InstantArticle;
 
 use Facebook\InstantArticles\Elements\Header;
+use Facebook\InstantArticles\Elements\Author;
 use Facebook\InstantArticles\Transformer\Rules\BoldRule;
 use Facebook\InstantArticles\Transformer\Rules\H1Rule;
 use Facebook\InstantArticles\Transformer\Rules\ItalicRule;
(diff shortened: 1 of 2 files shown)


everton-rosario:
@timjacobi Nice fix!
I have mixed feelings about this PR since it will conflict with #287. 
Into that particular PR I just change the properties `src`, `link` and `href`, which I was trying to be sure to be dealing with only URLs/string.

But then, this PR is replacing any &amp; back to & in cases that maybe we shouldn't. Example, if we are transforming any `&amp;` into `&` this case will break:

```
Paragraph::create()->appendText("This is how e can use &amp; tags into normal paragraphs")->render();
```

What do you think @timjacobi?
Do you think we should merge these 2 PRs into one?
But then we can check for adding more test cases for rendering and also transforming. I see the transforming testCases more important in this case, in a sense that any parsed content, should be transformed correctly.

timjacobi:
I don't think that the two PRs will conflict but yours is definitely more important. Since we've both been good citizen and added tests I suggest we merge your PR first and then see if my PR rebased onto yours breaks any tests.

pestevez:
Hey, @timjacobi - can you please resolve the conflicts so we can take another look at the PR? Thanks!

vkama:
hey guys, I am jumping on this late. So #287 was merged and Tim rebased recently and the tests are passing, this means we are good to merge or should I do a more in depth test?

timjacobi:
It's good to go.

tests/Facebook/InstantArticles/Transformer/TransformerTest.php
```diff
@@ -34,6 +35,13 @@ public function testTransformString()
         $this->assertEqualsHtml('<h1>Title String</h1>', $header->getTitle()->render());
     }
 
+    public function testDontTransformHTMLEntitiesTwice()
+    {
+        $author = Author::create()->withName("Test &amp; Test");
```
everton-rosario: We need another test case here, just to be sure we are doing it correctly with the &amp;

`Author::create()->withName("Test & Test")`
and
`Author::create()->withName("Test & Test")->withURL('http://facebook.com/param1=val1&param2=val2')`

vkama: Ok I will merge this and create an issue to add the tests

vkama: #297

PR #279, opened 2017-11-13
Adds property to global rules for header Image

yeah-sir:
This PR adds a global rule for header image

* Added a property for header image in GlobalRule
* Updated tests for GlobalRules

Follows #.
Related to #.
Fixes #.

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
+++ b/src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
@@ -140,18 +140,18 @@ class GlobalRule extends ConfigurationSelectorRule
         }
 
         // Treats Header Image
-        $articleHeaderImage = $this->getProperty(self::PROPERTY_GLOBAL_HEADER_IMAGE, $node);
-        if ($articleHeaderImage) {
-            $header->withCover($articleHeaderImage);
+        $articleHeaderImageURL = $this->getProperty(self::PROPERTY_GLOBAL_HEADER_IMAGE, $node);
+        if ($articleHeaderImageURL) {
+            $header->withCover(Image::create()->withURL($articleHeaderImageURL));
         } else {
             $transformer->addWarning(
                 new InvalidSelector(
                 self::PROPERTY_GLOBAL_HEADER_IMAGE,
                 $instantArticle,
                 $node,
-                $this  
+                $this
               )
-            )
+            );
         }
 
         $body = $this->getProperty(self::PROPERTY_GLOBAL_BODY, $node);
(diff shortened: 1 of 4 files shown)


src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
```diff
@@ -136,6 +139,21 @@ public function apply($transformer, $instantArticle, $node)
             $header->withTime(Time::create(Time::PUBLISHED)->withDatetime($timePublished));
         }
 
+        // Treats Header Image
+        $articleHeaderImageURL = $this->getProperty(self::PROPERTY_GLOBAL_HEADER_IMAGE, $node);
+        if ($articleHeaderImageURL) {
+            $header->withCover(Image::create()->withURL($articleHeaderImageURL));
+        } else {
```
diegoquinteiro: Maybe we don't need to add a warning here as the cover image is optional?

yeah-sir: Makes sense. I'll remove that block of code

src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
```diff
@@ -136,6 +139,12 @@ public function apply($transformer, $instantArticle, $node)
             $header->withTime(Time::create(Time::PUBLISHED)->withDatetime($timePublished));
         }
 
+        // Treats Header Image
+        $articleHeaderImageURL = $this->getProperty(self::PROPERTY_GLOBAL_HEADER_IMAGE, $node);
+        if ($articleHeaderImageURL) {
+            $header->withCover(Image::create()->withURL($articleHeaderImageURL));
```
everton-rosario: If this is the only way we are adding the image to the header, then, we are missing the opportunity to add caption content to this image as well.

Probably you will need to add a call to `$transformer->transform($header, $node)` or to find the caption by some other property you set on the configuration of the GlobalRule as well (some sort of `image.caption` property name to be created).

tests/Facebook/InstantArticles/Transformer/GlobalHtml/global.html
```diff
@@ -14,6 +14,9 @@ <h1>The <b>article</b> title</h1>
                 <img src="http://domain.com/image-body.png"/>
                 <span class="caption">The caption for image.</span>
             </div>
+            <div class="hero-image">
```
everton-rosario: Consider adding a caption block of text to be added to the image on the header.
This way we will cover not only images, but also caption to these images.

src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
yeah-sir: Thanks @everton-rosario. I had spoken with @diegoquinteiro about this and he mentioned that image caption was low priority for now. @diegoquinteiro can you please confirm?

If we are considering image caption along with this change I will make sure make the appropriate code changes.

diegoquinteiro: @yeah-sir confirmed! Let's merge this one and follow up with an issue.

PR #280, opened 2017-11-16
Hhvm strict

everton-rosario:
This PR

* [x] Transforms code into strict HHVM
* [x] Adds more tests to cover 
* [x] Fixes whatever needed to


src/Facebook/InstantArticles/Elements/Caption.php
```diff
@@ -345,62 +317,60 @@ public function getPosition()
...
     /**
-     * Structure and create the full ArticleImage in a XML format DOMElement.
+     * Structure and create the full ArticleImage in a XML format DOMNode.
      *
      * @param \DOMDocument $document where this element will be appended. Optional
-     * @return \DOMElement
+     * @return \DOMNode
      */
-    public function toDOMElement($document = null)
+    public function toDOMElement(\DOMDocument $document): \DOMNode
```
timjacobi: `$document` was optional before. Should it stay this way?
I think `toDOMElement(\DOMDocument $document = new \DOMDocument())` would have the same semantics.

src/Facebook/InstantArticles/Elements/Cite.php
```diff
@@ -120,51 +109,47 @@ public function withPostion($position)
...
-     * Structure and create the <cite> in a DOMElement.
+     * Structure and create the <cite> in a DOMNode.
      *
      * @param \DOMDocument $document - The document where this element will be appended (optional).
      *
-     * @return \DOMElement
+     * @return \DOMNode
      */
-    public function toDOMElement($document = null)
+    public function toDOMElement(\DOMDocument $document): \DOMNode
```
timjacobi: Same as above

src/Facebook/InstantArticles/Elements/Div.php
```diff
@@ -32,14 +32,10 @@ public static function create()
      *
      * @param \DOMDocument $document - The document where this element will be appended (optional).
      *
-     * @return \DOMElement
```
timjacobi: Are `DOMElement` and `DOMNode` the same?

src/Facebook/InstantArticles/Transformer/Getters/JSONGetter.php
```diff
@@ -10,45 +10,27 @@
...
             $this->withAttribute($properties['attribute']);
         }
-    }
-
-    /**
-     * @param string $attribute
-     *
-     * @return $this
-     */
-    public function withAttribute($attribute)
```
timjacobi: Has this moved somewhere else?

src/Facebook/InstantArticles/Elements/Caption.php
```diff
@@ -345,62 +317,60 @@ public function getPosition()
...
 
         $element = $document->createElement('figcaption');
 
         // title markup REQUIRED
-        if ($this->title && (!$this->subTitle && !$this->credit)) {
-            $element->appendChild($this->title->textToDOMDocumentFragment($document));
-        } elseif ($this->title) {
-            $element->appendChild($this->title->toDOMElement($document));
+        if ($this->title !== null) {
+            if (!$this->subTitle && !$this->credit) {
```
diegoquinteiro: Lets swap the if/else conditions and add some comments.

```php
if ($this->subTitle != null || $this->credit != null) {
    // Use the <h1> node if a sub title or credit is present
} else {
    // Use a simple loose text node if the caption is by itself
}
```

src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
```diff
@@ -10,50 +10,50 @@
...
+class DateGetter extends AbstractGetter
 {
     /**
      * @var string
      */
-    protected $format;
+    protected ?string $format;
 
-    public function createFrom($properties)
+    public function createFrom(array<string, string> $properties): this
```
diegoquinteiro: Should we use `Map` instead of `array`? Either would work, but we're already using Hack collections in other places. No strong feelings, it's your call.

src/Facebook/InstantArticles/Transformer/Getters/GetterFactory.php
```diff
@@ -43,13 +43,13 @@ class GetterFactory
      * @see JSONGetter
      * @see XpathGetter
      *
-     * @param string[] $getter_configuration that maps the properties for getter
+     * @param Map<string, string> $getter_configuration that maps the properties for getter
      *
      * @return AbstractGetter
      */
-    public static function create($getter_configuration)
+    public static function create(array<string, mixed> $getter_configuration): AbstractGetter
```
diegoquinteiro: Here your comment says `Map`, but the code says `array`. Once we decide which one to use, please update the comments accordingly

src/Facebook/InstantArticles/Transformer/Getters/JSONGetter.php
diegoquinteiro: This method already exists on `AbstractGetter`, so after the refactor you can remove it.

src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
```diff
@@ -153,16 +159,93 @@ public function findAll($node, $selector)
      * @param $node
      * @return null
      */
-    public function getProperty($property_name, $node)
+    public function getProperty(string $property_name, \DOMNode $node): mixed
```
diegoquinteiro: The comment says `@return null`. Let's fix the comment.

src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
```diff
@@ -153,16 +159,93 @@ public function findAll($node, $selector)
...
         return $value;
     }
 
-    public function getProperties()
+    /**
+     * @param $property_name
+     * @param $node
+     * @return the string
+     */
+    public function getPropertyString(string $property_name, \DOMNode $node): ?string
```
diegoquinteiro: Let's open an issue to retire the "type" parameter on the getter configuration? We're already forcing specific return types from the getter, so we could at least make it optional.

src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
```diff
@@ -26,67 +29,69 @@ class GlobalRule extends ConfigurationSelectorRule
...
                 self::PROPERTY_GLOBAL_AUTHOR_NAME,
                 self::PROPERTY_GLOBAL_AUTHOR_DESCRIPTION,
                 self::PROPERTY_GLOBAL_AUTHOR_ROLE_CONTRIBUTION,
                 self::PROPERTY_GLOBAL_CANONICAL_URL,
                 self::PROPERTY_GLOBAL_TITLE,
                 self::PROPERTY_TIME_PUBLISHED,
-                self::PROPERTY_GLOBAL_BODY
-            ],
-            $properties
+                self::PROPERTY_GLOBAL_BODY,
```
diegoquinteiro: Let's make sure we merge #279 into the hhvm branch as well.

src/Facebook/InstantArticles/Transformer/Rules/HeaderKickerRule.php
```diff
@@ -8,28 +8,37 @@
...
+        $kickerRule->withSelector(Type::mixedToString($configuration['selector']));
+        return $kickerRule;
     }
 
-    public function apply($transformer, $header, $h3)
+    public function apply(Transformer $transformer, Element $header, \DOMNode $h3): Element
     {
-        $header->withKicker($h3->textContent);
+        invariant($header instanceof Header, 'Error, $header is not Header');
+        $kicker = $transformer->transform(H3::create(), $h3);
```
diegoquinteiro: You don't need the invariant here. You can refactor to:

```php
$kicker = H3:create();
$transformer->transform($kicker, $h3);
```

src/Facebook/InstantArticles/Transformer/Rules/HeaderSubTitleRule.php
```diff
@@ -8,29 +8,37 @@
...
+        $headerSubTitleRule->withSelector(Type::mixedToString($configuration['selector']));
+        return $headerSubTitleRule;
     }
 
-    public function apply($transformer, $header, $h2)
+    public function apply(Transformer $transformer, Element $header, \DOMNode $h2): Element
     {
-        $header->withSubTitle($transformer->transform(H2::create(), $h2));
+        invariant($header instanceof Header, 'Error, $header not Header');
+        $subtitle = $transformer->transform(H2::create(), $h2);
```
diegoquinteiro: Same as H3 here;

src/Facebook/InstantArticles/Transformer/Rules/HeaderTitleRule.php
```diff
@@ -8,29 +8,37 @@
...
+        return $headerTitleRule;
     }
 
-    public function apply($transformer, $header, $h1)
+    public function apply(Transformer $transformer, Element $header, \DOMNode $h1): Element
     {
-        $header->withTitle($transformer->transform(H1::create(), $h1));
+        invariant($header instanceof Header, 'Error, $header is not Header');
+        $title = $transformer->transform(H1::create(), $h1);
+        invariant($title instanceof H1, 'Error, $title is not H1');
```
diegoquinteiro: Same here too.

src/Facebook/InstantArticles/Transformer/Rules/InteractiveInsideParagraphRule.php
```diff
@@ -1,4 +1,4 @@
-<?hh //decl
+<?hh // strict
```
diegoquinteiro: Hey, this rule was deprecated, since Interactive accepts Paragraph as a context. Why is it back?

src/Facebook/InstantArticles/Transformer/Rules/ParagraphFooterRule.php
```diff
@@ -8,36 +8,37 @@
  */
 namespace Facebook\InstantArticles\Transformer\Rules;
 
+use Facebook\InstantArticles\Elements\Element;
 use Facebook\InstantArticles\Elements\Paragraph;
 use Facebook\InstantArticles\Elements\Footer;
+use Facebook\InstantArticles\Validators\Type;
+use Facebook\InstantArticles\Transformer\Transformer;
 
 class ParagraphFooterRule extends ConfigurationSelectorRule
```
diegoquinteiro: We can probably deprecate this rule as well if we add Footer as a possible context for Paragraph.

src/Facebook/InstantArticles/Transformer/Rules/Rule.php
```diff
@@ -8,72 +8,87 @@
...
-        if ($node->nodeName === 'iframe') {
-            $log->debug('context class: '.get_class($context));
-            $log->debug('context matches: '.($matches_context ? 'MATCHES' : 'no match'));
-            $log->debug('node name: <'.$node->nodeName.' />');
-            $log->debug('node: '.$node->ownerDocument->saveXML($node).' />');
-            $log->debug('node matches: '.($matches_node ? 'MATCHES' : 'no match'));
-            $log->debug('rule: '.get_class($this));
-            $log->debug('-------');
-        }
+        // if ($matches_context && $matches_node) {
```
diegoquinteiro: Let's get rid of all these commented lines =)

diegoquinteiro: Let's also create an issue for adding proper verbosity levels.

src/Facebook/InstantArticles/Transformer/Rules/Rule.php
```diff
@@ -8,72 +8,87 @@
...
+    {
+        return get_called_class();
+    }
+
+    /**
+     * Auxiliary method to extract all Elements full qualified class name.
+     *
+     * @return string The full qualified name of class.
+     */
+    public function getObjClassName(): string
```
diegoquinteiro: How is that different from `getClassName()`? If it's a leftover, let's get rid of it.

src/Facebook/InstantArticles/Transformer/Transformer.php
```diff
@@ -8,48 +8,40 @@
...
 use Facebook\InstantArticles\Validators\InstantArticleValidator;
 
 class Transformer
 {
     /**
-     * @var Rule[]
+     * @var array<array<int, Rule>> This is the internal map for rules to be applied
      */
-    private $rules = [];
+    private array<Rule> $rules = array();
```
diegoquinteiro: Let's do a project wide search for `array<` to review outdated documentation.

src/Facebook/InstantArticles/Transformer/Transformer.php
```diff
@@ -97,82 +91,42 @@ public static function markAsProcessed($node)
      *
      * @param DOMNode $node The node to clone
```
diegoquinteiro: Outdated docs as well, let's change it to:
```php
/**
 * @param DOMNode The node of interest
 * @return bool Whether the node is processed or not
 */
```

src/Facebook/InstantArticles/Transformer/Transformer.php
```diff
@@ -231,27 +185,12 @@ public function transform($context, $node)
                 }
                 $matched = false;
 
-                // Get all classes and interfaces this context extends/implements
```
diegoquinteiro: Why are we removing the performance optimization? The algorithm is 10 times slower without it.

diegoquinteiro: The changes you reverted on addRule, getRules and getAllClassTypes were all related to this optimization.

diegoquinteiro: Here's the original perf PR for reference: #46

src/Facebook/InstantArticles/Transformer/Rules/InteractiveInsideParagraphRule.php
diegoquinteiro: Here's the original PR for reference #186

src/Facebook/InstantArticles/Client/Helper.php
```diff
@@ -8,26 +8,24 @@
...
 namespace Facebook\InstantArticles\Client;
 
-use Facebook\Exceptions\FacebookResponseException;
-use Facebook\Facebook;
-use Facebook\Exceptions\FacebookSDKException;
-use Facebook\Authentication\AccessToken;
+use \Facebook\Exceptions\FacebookResponseException;
+use \Facebook\Facebook;
+use \Facebook\Exceptions\FacebookSDKException;
+use \Facebook\Authentication\AccessToken;
```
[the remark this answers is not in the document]
everton-rosario: The conflict we have here is just because our namespace also starts with Facebook.
But then, to the point of reconfiguring the autogen/class loader, I couldn't get to a place where I'd have it working for both scenarios.

The /client was dropped from internal version (not needed) then, this was the only place I used for used code.

Also only classes from the standard API from hhvm/php is using, just like: `\Map`, `\DOMNode`, `\DOMDocument` etc...

src/Facebook/InstantArticles/Elements/Caption.php
everton-rosario: Technically the main issue of us rendering the elements and creating by ourselves the DOMDocument, is that we will miss the correct encoding to write content to.
I feel that keeping it optional (so we create it all the time) works well only for test purposes, but for practical usage of our lib, we should drop this being optional.

What do you think?

src/Facebook/InstantArticles/Elements/Cite.php
everton-rosario: https://github.com/facebook/facebook-instant-articles-sdk-php/pull/280#discussion_r151810807

src/Facebook/InstantArticles/Elements/Div.php
everton-rosario: Technically all DOMElement is a DOMNode, but not the other way around.

While parsing a DOMDocument and navigating through the children elements, only standard tags will map to DOMElement.

src/Facebook/InstantArticles/Transformer/Getters/JSONGetter.php
everton-rosario: As @diegoquinteiro says.
Removed for no needing to keep it here.

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -22,29 +21,25 @@ class Client
     /**
      * @var Facebook The main Facebook service client.
      */
-    private $facebook;
+    private \Facebook\Facebook $facebook;
```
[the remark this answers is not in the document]
everton-rosario: Namespace conflicts for autoload =(

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -22,29 +21,25 @@ class Client
...
 
     /**
-     * @var int ID of the Facebook Page we are using for Instant Articles
+     * @var string ID of the Facebook Page we are using for Instant Articles
      */
-    protected $pageID;
+    protected string $pageID;
 
     /**
      * @var bool|false Are we using the Instant Articles development sandbox?
```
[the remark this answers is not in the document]
everton-rosario: yes, changing.

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -64,20 +59,16 @@ public function __construct($facebook, $pageID, $developmentMode = false)
...
-
         $facebook = new Facebook([
             'app_id' => $appID,
             'app_secret' => $appSecret,
             'default_access_token' => $accessToken,
             'default_graph_version' => 'v2.5'
         ]);
 
-        return new static($facebook, $pageID, $developmentMode);
+        return new self($facebook, $pageID, $developmentMode);
```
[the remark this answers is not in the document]
everton-rosario: For strict code, this is the error:

`Can't use new static() for Facebook\InstantArticles\Client\Client; __construct arguments are not guaranteed to be consistent in child classes
This declaration neither defines an abstract/final __construct nor uses <<__ConsistentConstruct>> attribute
: src/Facebook/InstantArticles/Client/Client.php:17`

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -89,17 +80,14 @@ public static function create($appID, $appSecret, $accessToken, $pageID, $develo
...
-        Type::enforce($article, 'Facebook\InstantArticles\Elements\InstantArticleInterface');
-        Type::enforce($published, Type::BOOLEAN);
-
         // Never try to take live if we're in development (the API would throw an error if we tried)
         $published = $this->developmentMode ? false : $published;
 
         // Assume default access token is set on $this->facebook
         $response = $this->facebook->post($this->pageID . Client::EDGE_NAME, [
-          'html_source' => $article->render(null, $formatOutput),
+          'html_source' => $article->render(),
```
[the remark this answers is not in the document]
everton-rosario: Filling a issue so we treat correctly the charset encoding.
We don't have a good formatter into HHVM also, so we will need to fix this as well.

2 new issues: #281 #282

src/Facebook/InstantArticles/Client/Client.php
```diff
@@ -157,17 +143,15 @@ public function removeArticle($canonicalURL)
...
-        Type::enforce($canonicalURL, Type::STRING);
-
         $field = $this->developmentMode ? 'development_instant_article' : 'instant_article';
 
         $response = $this->facebook->get('?id=' . $canonicalURL . '&fields=' . $field);
         $instantArticle = $response->getGraphNode()->getField($field);
 
         if (!$instantArticle) {
-            return null;
+            return -1;
```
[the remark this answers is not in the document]
everton-rosario: Nice catch.
Fixing.

src/Facebook/InstantArticles/Client/ClientException.php
```diff
@@ -1,4 +1,4 @@
-<?hh //decl
+<?hh // strict
```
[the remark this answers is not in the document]
everton-rosario: Marking //strict means that all the code and all the code it refers to should be hhvm too.

For particular cases of using on Client (graph-api-sdk == php)  and 2 classes: DOMCdataSection and CSSSelector from symphony (are not hhvm strict)

src/Facebook/InstantArticles/Client/InstantArticleStatus.php
```diff
@@ -18,140 +18,135 @@ class InstantArticleStatus
...
+    /**
+     * @var Vector<ServerMessage>
+     */
+    private Vector<ServerMessage> $messages = Vector {};
+
     /**
      * @var ServerMessage[]
      */
-    private $messages = [];
+    private string $status = "";
```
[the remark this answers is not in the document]
everton-rosario: Nice catch! 💅

src/Facebook/InstantArticles/Elements/Ad.php
```diff
@@ -149,24 +143,20 @@ public function getSource()
...
-     * Structure and create the full Ad in a DOMElement.
+     * Structure and create the full Ad in a DOMNode.
      *
      * @param \DOMDocument $document - The document where this element will be appended (optional).
      *
-     * @return \DOMElement
+     * @return \DOMNode
      */
-    public function toDOMElement($document = null)
+    public function toDOMElement(\DOMDocument $document): \DOMNode
```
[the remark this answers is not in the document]
everton-rosario: Yes, lets add this to these future issues: #281 #282

src/Facebook/InstantArticles/Elements/Caption.php
everton-rosario: Good catch!
Fixing

src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
everton-rosario: The first find/replace generated tons of bugs, so moved back to array.
It would be better to use Map, for sure.

src/Facebook/InstantArticles/Transformer/Getters/GetterFactory.php
everton-rosario: Fixing to match code

src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
everton-rosario: Fixed! Nice catch!

src/Facebook/InstantArticles/Transformer/Rules/ConfigurationSelectorRule.php
everton-rosario: Yes, agreed. #283

src/Facebook/InstantArticles/Transformer/Rules/GlobalRule.php
everton-rosario: Agreed.

src/Facebook/InstantArticles/Transformer/Rules/HeaderKickerRule.php
everton-rosario: Good call!

src/Facebook/InstantArticles/Transformer/Rules/HeaderSubTitleRule.php
everton-rosario: Copy/paste effects!
Good call!

src/Facebook/InstantArticles/Transformer/Rules/HeaderTitleRule.php
everton-rosario: Copy/paste effects!
Good call!

src/Facebook/InstantArticles/Transformer/Rules/InteractiveInsideParagraphRule.php
everton-rosario: Class was still there, but not being used, indeed.

everton-rosario: Issue created: #284

src/Facebook/InstantArticles/Transformer/Rules/Rule.php
everton-rosario: Indeed: #285

src/Facebook/InstantArticles/Transformer/Rules/ParagraphFooterRule.php
everton-rosario: Indeed: #286

src/Facebook/InstantArticles/Transformer/Rules/Rule.php
everton-rosario: Unfortunately using //strict it forces to call each method properly.
so self::getClassName() and $instance->getObjClassName() are required for code =(

src/Facebook/InstantArticles/Transformer/Transformer.php
everton-rosario: 👍 
Fixed know ones.

src/Facebook/InstantArticles/Transformer/Transformer.php
everton-rosario: Fixed.

src/Facebook/InstantArticles/Transformer/Transformer.php
everton-rosario: This is one of the discussion points.
This was all broken(after strict migration), not well tested and documented. Got rid of it.
We can add it back with proper review.

src/Facebook/InstantArticles/Elements/Caption.php
timjacobi: Sounds good

diegoquinteiro:
:shipit:

PR #289, opened 2017-11-24
Fix: Remove parentheses

localheinz:
This PR

* [x] removes unnecessary parentheses


localheinz:
Thank you, @everton-rosario!

PR #304, opened 2018-01-12
Implemented logging structure

### LINKED ISSUE #240: Transformer error logging
# Steps required to reproduce the problem

1. Call `Facebook\InstantArticles\Parser::parse`

# Expected Result

I should receive instance of instant article if transformation worked. 
In case it failed I should receive exception or array of exceptions. 

# Actual Result

All debug messages are logged to output.

### LINKED ISSUE #245: Using composer - error on log4php
# Steps required to reproduce the problem

1. Using Drupal 7 and Composer Manager, enabled Facebook Instant Articles module
2. composer.json is set to rebuild automatically - confirmed that FIA requirements are in file
3. php composer.phar install

# Expected Result

* installation of all dependencies in /sites/all/modules

# Actual Result

* Failure.  Here's the error:

`[RuntimeException]                                                                                                  
  Failed to execute git clone --no-checkout 'https://git-wip-us.apache.org/repos/asf/logging-log4php.git' 'myvendordirectorylocation/vendor/apache/log4php' --dissociate --reference 'mycomposerdirectorylocation/.composer/cache/  
  vcs/https---git-wip-us.apache.org-repos-asf-logging-log4php.git/' && cd '/myvendordirectorylocation/vend  
  or/apache/log4php' && git remote add composer 'https://git-wip-us.apache.org/repos/asf/logging-log4php.git' && git  
   fetch composer                                                                                                     
  Cloning into '/myvendordirectorylocation/vendor/apache/log4php'...                                        
  fatal: The remote end hung up unexpectedly                                                                          
                                             `

### LINKED ISSUE #302: Composer install log4php problem
I try to install with composer and it gives this error : 

```
  [RuntimeException]
  Failed to clone https://git-wip-us.apache.org/repos/asf/logging-log4php.git, git was not found, check that 
  it is in  stalled and in your PATH env.  'git' is not recognized as an internal or external command,
  operable program or batch file.
```

everton-rosario:
* [x] Implemented a simple internal logging structure.
* [x] Removed log4php
* [x] Fixed #302
* [x] Fixed #245
* [x] Fixed #240

Calls to Transformer have now stored all logging messages:
```
        $json_file = file_get_contents(__DIR__ . '/simple-rules.json');

        $instant_article = InstantArticle::create();
        $transformer = new Transformer();
        $transformer->loadRules($json_file);

        $html_file = file_get_contents(__DIR__ . '/simple.html');

        TransformerLog::setLevel(TransformerLog::DEBUG);
        $transformer->transformString($instant_article, $html_file);
        $result = $transformer->getLogs();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ If you are encountering problems, the following tips may help in troubleshooting
 
 - If content is missing from your transformed article, more likely than not there isn't a **Transformer Rule** matching an element in your source markup. See how to configure appropriate rules for your content in the [Transformer Rules documentation](https://developers.facebook.com/docs/instant-articles/sdk/transformer-rules).
 
-- Set the `threshold` in the [configuration of the Logger](https://logging.apache.org/log4php/docs/configuration.html#PHP) to `DEBUG` to expose more details about the items processed by the Transformer.
+- Set the `threshold` in the transformer Logger level to `DEBUG` to expose more details about the items processed by the Transformer. `$transformer->setLogLevel(Transformer::LOG_DEBUG);`
 
 - Refer to the existing [tests of the `Elements`](https://github.com/facebook/facebook-instant-articles-sdk-php/tree/master/tests/Facebook/InstantArticles/Elements) for examples of what is required of each and to potentially create your own tests (which can be run with `$ composer test`).
 
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ If you are encountering problems, the following tips may help in troubleshooting
 
 - If content is missing from your transformed article, more likely than not there isn't a **Transformer Rule** matching an element in your source markup. See how to configure appropriate rules for your content in the [Transformer Rules documentation](https://developers.facebook.com/docs/instant-articles/sdk/transformer-rules).
 
-- Set the `threshold` in the transformer Logger level to `DEBUG` to expose more details about the items processed by the Transformer. `$transformer->setLogLevel(Transformer::LOG_DEBUG);`
+- Set the `threshold` in the transformer Logger level to `DEBUG` to expose more details about the items processed by the Transformer. `TransformerLog::setLevel(TransformerLog::DEBUG);`
 
 - Refer to the existing [tests of the `Elements`](https://github.com/facebook/facebook-instant-articles-sdk-php/tree/master/tests/Facebook/InstantArticles/Elements) for examples of what is required of each and to potentially create your own tests (which can be run with `$ composer test`).
 
--- a/composer.json
+++ b/composer.json
@@ -11,13 +11,14 @@
     }],
     "require": {
         "php": "^5.4 || ^7.0",
-        "symfony/css-selector": "2.8.* || ^3.0",
-        "facebook/graph-sdk": "~5.0",
-        "apache/log4php": "2.3.0"
+        "symfony/css-selector": "2.8.*",
+        "facebook/graph-sdk": "~5.0"
     },
     "require-dev": {
         "fzaninotto/faker": "^1.6.0",
-        "phpunit/phpunit": "^4.8.36",
+        "phpunit/phpunit": "4.8.*",
+        "symfony/yaml": "2.1.*",
+        "phpdocumentor/reflection-docblock": "^2.0",
         "squizlabs/php_codesniffer": "^2.6.0"
     },
     "autoload": {
--- a/src/Facebook/InstantArticles/Client/ServerMessage.php
+++ b/src/Facebook/InstantArticles/Client/ServerMessage.php
@@ -73,8 +73,6 @@ class ServerMessage
         if ($validLevel) {
             return new self($level, $message);
         } else {
-            \Logger::getLogger('facebook-instantarticles-client')
-                ->info('Unknown message level "$level". Are you using the last SDK version?');
             return new self(self::INFO, $message);
         }
     }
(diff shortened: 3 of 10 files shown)


composer.json
```diff
@@ -11,9 +11,8 @@
     }],
     "require": {
         "php": "^5.4 || ^7.0",
-        "symfony/css-selector": "2.8.* || ^3.0",
-        "facebook/graph-sdk": "~5.0",
-        "apache/log4php": "2.3.0"
+        "symfony/css-selector": "2.8.*",
```
Jayad: Can we upgrade this to latest one(i guess 4.0.3) if no dependency with this version(2.8.*)?

src/Facebook/InstantArticles/Transformer/Transformer.php
```diff
@@ -57,6 +57,43 @@ class Transformer
...
+
+    /**
+     * @var string $logLevel The log level set into this transformer instance. Possible values: LOG_NOLOG, LOG_INFO or LOG_DEBUG
+     */
+    private $logLevel = self::LOG_INFO;
+
+    /**
+     * @var string[] $logs The log messages generated during the transformation process.
+     */
+    private $logs = ['Possible log levels: OFF, ERROR, INFO or DEBUG'];
```
Jayad: How about WARN or FATAL?

README.md
```diff
@@ -82,7 +82,7 @@ If you are encountering problems, the following tips may help in troubleshooting
 
 - If content is missing from your transformed article, more likely than not there isn't a **Transformer Rule** matching an element in your source markup. See how to configure appropriate rules for your content in the [Transformer Rules documentation](https://developers.facebook.com/docs/instant-articles/sdk/transformer-rules).
 
-- Set the `threshold` in the [configuration of the Logger](https://logging.apache.org/log4php/docs/configuration.html#PHP) to `DEBUG` to expose more details about the items processed by the Transformer.
+- Set the `threshold` in the transformer Logger level to `DEBUG` to expose more details about the items processed by the Transformer. `$transformer->setLogLevel(Transformer::LOG_DEBUG);`
```
Jayad: Just Curious, What will be the advantage of using Logger implemented within the same code, instead of using the log4php library which already has lot of functionality.

src/Facebook/InstantArticles/Client/ServerMessage.php
```diff
@@ -73,8 +73,6 @@ public static function fromLevel($level, $message)
         if ($validLevel) {
             return new self($level, $message);
         } else {
-            \Logger::getLogger('facebook-instantarticles-client')
-                ->info('Unknown message level "$level". Are you using the last SDK version?');
```
Jayad: Will this file functions will be used over Transformer.php implementations?

composer.json
everton-rosario: Anyone is able to upgrade it, but we cannot move to 3.2 and up since it is not php 5.4 compatible.
We need to re-think the need of keeping php 5.4 compatibility

src/Facebook/InstantArticles/Client/ServerMessage.php
everton-rosario: No.
All the package `/Client` is on SDK for making it easier to call all APIs for the IA ingestion. [See more here](https://developers.facebook.com/docs/instant-articles/api).

src/Facebook/InstantArticles/Transformer/Transformer.php
everton-rosario: Warnings are all ready to be accessed from `$transformer->getWarnings()`.

Fatals I see as the Exception handling being the best approach.

everton-rosario:
Hey @mburak , good call on moving it away from Transformer class.

Check the new commit with this in place: https://github.com/facebook/facebook-instant-articles-sdk-php/pull/304/commits/e954e7ed04362cffe2702812bc8554b6eea93bf0

mburak:
hey @everton-rosario, it still looks a little bit coupled for me. What if instead of having the logs being added (that logic) in the Tranformer class we do that in the TransformerLog class (and maybe TranformerLogger is a better naming?).

So instead of having this:
`$this->addLog(
             TransformerLog::INFO,
             'Possible log levels: OFF, ERROR, INFO or DEBUG. To change it call method TransformerLog::setLevel("DEBUG").'
         );` 

we could have something like this:
`TransformerLogger::logInfo($this->logs,
             'Possible log levels: OFF, ERROR, INFO or DEBUG. To change it call method TransformerLog::setLevel("DEBUG").'
         );`

so in this case we leave all the logging logic on the TransformerLogger class. I'm also adding an abstraction here so we don't use the different enums for each level and we just have a different method for each level logging. 
What do you think?

everton-rosario:
@mburak This is not a Logger tool. It was intended and built only for transformer, thats why I coupled it to the Transformer instance. 
If we handle it as a Logger tool, we will have all transformations altogether into same log.

Here, with this simple structure, we have a log structure that isolates only transformation debug info, to improve depuration on the transformer.

Since it has all logging about transformation only, users of the SDK can optionally use their own logging system to output the transformation Log that occurred if they intend to, leaving it as optional.

Also I don't see this logging system being used to anything else but Transformation debug.

mburak:
@everton-rosario gotcha. We could still handle each transformation in different logs even when having a decoupled logger. You pass the log array to the logger so the transformer still keeps the log state. 
Anyway, I understand that's not the intention of this.  Go for it!

everton-rosario:
Issue filed so we can treat it evenly not only for the logs, but also for the warnings:
https://github.com/facebook/facebook-instant-articles-sdk-php/issues/306

PR #310, opened 2018-01-29
Now cloning elements only in the right rules (interactive, ads, analy…

mburak:
…tics)

This PR changes the logic in a way that now we only clone the embeds/iframe for only the right elements (Interactive, Ads, Analytics, SocialEmbeds). 

This fixes the xpath element evaluation that wasn't working when a rule had selectors with comma.


pestevez:
Hey @mburak, thanks for this! How hard will it be to add some unit tests?

mburak:
Hey @pestevez, @everton-rosario, I added tests for comma selectors and missing rules (ads, analytics). Let me know what do you think.

pestevez:
Just a small caveat: since the original issue was uncovered using the `GlobalRule` and this PR fixes it, can you please include a test where this rule is also used? I think tweaking any of the existing selectors in file `global-rules.json` would do it. Thanks again!

mburak:
Done!

diegoquinteiro:
Thanks for the regression tests!

PR #315, opened 2018-02-07
Added support for new dateTextContent attr in GlobalRule

mburak:
This PR

* adds support for a new attribute dateTextContent used in GlobalRule for dates that are in the content side and not in the datetime attribute.

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
+++ b/src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
@@ -50,8 +50,8 @@ class DateGetter extends StringGetter
         if (!empty($elements) && $elements->item(0)) {
             $element = $elements->item(0);
 
-            if ($this->attribute && $this->attribute != 'dateTextContent') {
-                return \DateTime::createFromFormat($this->format, $element->getAttribute($this->attribute));
+            if ($this->attribute) {
+                return \DateTime::createFromFormat('!'.$this->format, $element->getAttribute($this->attribute));
             }
             return \DateTime::createFromFormat('!'.$this->format, trim($element->textContent));
         }
--- a/src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
+++ b/src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
@@ -50,10 +50,10 @@ class DateGetter extends StringGetter
         if (!empty($elements) && $elements->item(0)) {
             $element = $elements->item(0);
 
-            if ($this->attribute) {
+            if ($this->attribute && $this->attribute != 'dateTextContent') {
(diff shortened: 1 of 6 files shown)


diegoquinteiro:
I don't clearly understand what this PR does. Can you give more context of how the rules will look like?

src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
```diff
@@ -50,10 +50,10 @@ public function get($node)
         if (!empty($elements) && $elements->item(0)) {
             $element = $elements->item(0);
 
-            if ($this->attribute) {
+            if ($this->attribute && $this->attribute != 'dateTextContent') {
                 return \DateTime::createFromFormat($this->format, $element->getAttribute($this->attribute));
             }
-            return \DateTime::createFromFormat($this->format, $element->textContent);
+            return \DateTime::createFromFormat('!'.$this->format, trim($element->textContent));
```
diegoquinteiro: Maybe that's the only line you should keep (plus the tests).

src/Facebook/InstantArticles/Transformer/Getters/DateGetter.php
```diff
@@ -50,10 +50,10 @@ public function get($node)
         if (!empty($elements) && $elements->item(0)) {
             $element = $elements->item(0);
 
-            if ($this->attribute) {
+            if ($this->attribute && $this->attribute != 'dateTextContent') {
```
diegoquinteiro: Let's not use a custom attribute name like 'dateTextContent'. We can simply omit the attribute on the rule to fetch the content, that's the default for all rules and getters.

tests/Facebook/InstantArticles/Transformer/GlobalHtml/global-rules-with-text-date.json
```diff
@@ -0,0 +1,221 @@
...
+                },
+                "article.canonical": {
+                    "type" : "string",
+                    "selector": "head meta",
+                    "attribute": "content"
+                },
+                "article.publish": {
+                    "selector": "div.date",
+                    "type": "date",
+                    "attribute": "dateTextContent",
```
diegoquinteiro: Just remove this line to accomplish the same behavior.

tests/Facebook/InstantArticles/Transformer/TransformerTest.php
```diff
@@ -92,8 +92,6 @@ public function testSelfTransformerContent()
         $instant_article->addMetaProperty('op:generator:transformer:version', '1.0.0');
         $result = $instant_article->render('', true)."\n";
 
-        //var_dump($result);
```
diegoquinteiro: 💅 👍

mburak:
fixes based on @diegoquinteiro comments

PR #318, opened 2018-02-14
Changes to allow Windows with php7(?) to work properly also with symfony 3.3.* projects

gurmiguel:
This PR
* assures the project to work properly with some specific systems where **mt_rand** won't work properly while testing and allow symfony 3.3.* projects to require it.


gurmiguel:
Invalid_

PR #319, opened 2018-02-20
Fix symfony requirement for css selector

alexander-schranz:
Fix problem with symfony 3

It was falsly removed by #304 because of thinking you need PHP 5.4 support also in the requirements but thats totally false! It will install the older dependency symfony 2 if you only have php 5.4.


alexander-schranz:
@mburak If I will generate a lock file it will be a lock file with library version of PHP7.1 thats why it don't make sense in a Library to commit the lock file as another developer with an older PHP Version will have problem with this lock file.

mburak:
@alexander-schranz Thanks for your recommendation but I don't think we should gitignore composer.lock since that goes against the recommendation from Composer and could potentially break our CI build.
You wouldn't use this SDK directly and it will be a dependency of other project, so the lock file won't have any effect on projects that depend on this one, it will still download the right dependencies, either you are using PHP 5.4 or 7.1

alexander-schranz:
@mburak the recommendation of composer is for projects where you only have one supported PHP Version this you use on production server. The lock file can't use in the CI with different PHP Versions as a lock file is PHP Version specific based on the Version who did create the lock file. And you want always run the CI on the newest version else you don't get an error when you are incompatible with newer version of some library you depend on. See any other PHP libraries which need to support multiple PHP Versions.

Just to make clear. If somebody install this library composer does not look into the composer.lock file of this library this is completely ignored by composer. So only the CI will use the composer.lock which will not work for supporting multiple PHP Version and its really not recommended as you will not test if you are really compatible with your requirements.

mburak:
@alexander-schranz currently we are using and testing effectively different PHP versions in our CI as you can see here: https://travis-ci.org/facebook/facebook-instant-articles-sdk-php and it's using a locked version of the composer as we need to keep that in a controlled environment for our tests.

alexander-schranz:
@mburak yeah but you are testing not agains the version constraint you did define in your composer.json.

mburak:
I'm not sure if I'm following you. We are testing against the 5.x and 7.x we support in the composer.json

mburak:
@alexander-schranz we'll give it a try the way you are posting it. Thanks for taking the time to work on and discuss this!

PR #349, opened 2019-03-20
Fix #346 - Transformer's transform method does not account for settings

### LINKED ISSUE #346: Transformer's transform method does not account for settings
# Steps required to reproduce the problem

1. Use a set of rules that include Analytics or Ads settings
2. Create instant article markup using the [Transformer::transform](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Transformer/Transformer.php#L285), instead of [Transformer::transformString](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Transformer/Transformer.php#L235)

# Expected Result
The markup includes the corresponding Analytics and/or Ads code

# Actual Result
The markup does not include the corresponding Analytics and/or Ads code because `transform` does not call `handleTransformationSettings`.

Note: To repro, replace `transformString` with `transform` in `SimpleTransformerTest::testSelfTransformerContent` and run the tests (`composer test`).

diegoquinteiro:
This PR

* [x] Fixes the `Transformer::transform` method to use the transformationSettings the same way as `Transformer::transformString`

Fixes #346

### DIFF
--- a/src/Facebook/InstantArticles/Transformer/Transformer.php
+++ b/src/Facebook/InstantArticles/Transformer/Transformer.php
@@ -263,9 +263,6 @@ class Transformer
         libxml_clear_errors();
         libxml_use_internal_errors($libxml_previous_state);
         $result = $this->transform($context, $document);
-        if (Type::is($result, InstantArticle::getClassName())) {
-            $result = $this->handleTransformationSettings($result);
-        }
         $totalTime = round(microtime(true) - $start, 3)*1000;
         $totalWarnings = count($this->getWarnings());
         $this->addLog(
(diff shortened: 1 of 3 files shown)


src/Facebook/InstantArticles/Transformer/Transformer.php
```diff
@@ -375,6 +374,10 @@ public function transform($context, $node)
             }
         }
 
+        if ($is_top_of_recursion_stack) {
+          $context = $this->handleTransformationSettings($context);
```
everton-rosario: Identation issue. Add 2 more spaces here. 💅

PR #354, opened 2020-07-31
Added a new h4 element

prilcool:
Added a new  h4 element in the elements folder

Follows #.
This class follows the original h1 and h2  element code structure .


pestevez:
Thanks for your contribution @prilcool. I might be wrong, but I don't think h4 is an element supported by the Instant Articles format.

According to [this doc](https://developers.facebook.com/docs/instant-articles/design/customize-design/), only h1 and h2 are supported.

PR #359, opened 2020-08-13
Drop unsupported PHP version in travis

ammarfaizi2:
Pre-installed version in xenial are only `5.6`, `7.1` and `7.2`.

https://docs.travis-ci.com/user/reference/xenial#php-support


TawanMiyaemphak:
ผม

PR #365, opened 2020-12-30
Move build to CircleCI

elomar:
This PR

* [x] moves the build from Travis to CircleCI
* [x] drops from the build versions that [WordPress doesn't recommend](https://wordpress.org/about/requirements/)

Related to https://github.com/Automattic/facebook-instant-articles-wp/pull/1069. Main difference from that is that we're unable to cache the dependencies due do the lock file being ignored.


pestevez (approved):
Looks great, thanks! I already enabled Circle CI and disabled Travis. 💯

PR #376, opened 2022-09-18
Fix strpos() deprecation notice

GaryJones:
PHP 8.1 generates a deprecation notice:

> strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated

This changes the `null` to an empty string if it hasn't been set.


GaryJones:
Closing for lack of interest.

